PR #2, opened 2017-08-25
Update java-lang boilerplate

willprice:


langs/java.go
```diff
@@ -124,8 +124,8 @@ func mavenOpts() string {
 	return opts.String()
 }
 
-/*	TODO temporarily generate maven project boilerplate from hardcoded values.
-	Will eventually move to using a maven archetype.
+/*    TODO temporarily generate maven project boilerplate from hardcoded values.
+Will eventually move to using a maven archetype.
 */
 
 const (
```
denismakogon: Startin thinking of extracting this into external XML, to make code cleaner. Thoughts?

willprice: Not averse to extracting it out, but i'd rather land this ASAP as our integration tests are broken since `cli` has gone from the main repo and the java support is currently broken in `fn` without these changes.

Ideally we'd be able to release and version lang support separately to the cli as it changes at a different pace to the cli code.

denismakogon: I'm not blocking this, already approved. Just a thought for future.

willprice: Sorry, I didn't mean to imply you were. I'll have a look at pulling out the XML this afternoon as I agree, it'd be nicer.

PR #3, opened 2017-08-25
Resubmit fn PR238 on cli

hhexo:
From fn PR238:

"Since I'm a Rust fan, in my spare time I'm contributing to the Rust build runtime for fn. :)

This change fixes the current build runtime (which is currently broken as funcy/rust:dev does not exist) and uses the official Rust image (which now exists on DockerHub), plus it generates boilerplate for a basic Cargo project."

### DIFF
--- a/langs/rust.go
+++ b/langs/rust.go
@@ -1,8 +1,10 @@
 package langs
 
 import (
+	"errors"
+	"io/ioutil"
 	"os"
-	"os/exec"
+	"path/filepath"
 )
 
 type RustLangHelper struct {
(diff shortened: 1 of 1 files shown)


langs/rust.go
```diff
@@ -1,43 +1,102 @@
...
 	BaseHelper
 }
 
 func (lh *RustLangHelper) BuildFromImage() string {
-	return "funcy/rust:dev"
+	return "rust:1"
 }
+
+func (lh *RustLangHelper) RunFromImage() string {
+	return "debian:stretch"
```
denismakogon: Tried to google for alpine version, because i think having huge images for functions is not the best case (bet we're gonna charge users for storage on Docker, hehe). So, maybe we can take a look at smaller things like https://hub.docker.com/r/frolvlad/alpine-rust/

hhexo: The problem with all the Alpine versions I found is that they don't keep up with Rust releases. Rust has a six-weeks release schedule with frequent updates, and since they finally have an official library image I hope they will keep that one up to date.

I have tried building on the official Rust image and running on Alpine but the binary fails to load - different libc, I think. So I looked up the official Rust image Dockerfile and found its base image.

I will try debian:stretch-slim to see if that works.

denismakogon: @hhexo any luck?

treeder: Maybe we can just merge it now and shrink it later.

denismakogon: I'm fine with this.

Issue #10, opened 2017-08-29
Routes update call overrider route attributes


denismakogon:
Existing route
```
fn routes i functions-hub /env
{
	"format": "http",
	"idle_timeout": 30,
	"image": "sample:0.0.4",
	"memory": 128,
	"path": "/env",
	"timeout": 50,
	"type": "sync"
}
```
Updating route:
```
fn routes u --ignore-fn-file --idle-timeout 50s functions-hub /env
functions-hub /env updated
```
Inspecting updated route
```
fn routes i functions-hub /env
{
	"format": "default",
	"idle_timeout": 50,
	"image": "sample:0.0.4",
	"memory": 128,
	"path": "/env",
	"timeout": 30,
	"type": "sync"
}
```

FN server works as expected:
```
fn routes i functions-hub /env
{
	"format": "default",
	"idle_timeout": 50,
	"image": "sample:0.0.4",
	"memory": 128,
	"path": "/env",
	"timeout": 30,
	"type": "sync"
}
```
```
curl -X PATCH localhost:8080/v1/apps/functions-hub/routes/env -d '{"route":{"timeout":50}}'
```
```
fn routes i functions-hub /env
{
	"format": "default",
	"idle_timeout": 50,
	"image": "sample:0.0.4",
	"memory": 128,
	"path": "/env",
	"timeout": 50,
	"type": "sync"
}
```

Issue #9, opened 2017-08-29
Deploy does not pick up idle timeout, headers, config from func.yaml


denismakogon:
CLI version 0.3.72

denismakogon:
It seems like `fn routes update` also doesn't do this.

PR #15, opened 2017-08-29
Fix route file and CLI arg setters

denismakogon:
Fixing file and CLI args setters that were overriding each other

Closes: #9 
Closes: #10

### DIFF
--- a/routes.go
+++ b/routes.go
@@ -221,43 +221,52 @@ func (a *routesCmd) call(c *cli.Context) error {
 }
 
 func routeWithFlags(c *cli.Context, rt *fnmodels.Route) {
-	if i := c.String("image"); i != "" {
-		rt.Image = i
-	}
-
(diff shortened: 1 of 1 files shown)


routes.go
```diff
@@ -221,43 +221,52 @@ func (a *routesCmd) call(c *cli.Context) error {
...
+			to := int32(t.Seconds())
+			rt.Timeout = &to
+		}
 	}
-
-	if t := c.Duration("idle-timeout"); t > 0 {
-		to := int32(t.Seconds())
-		rt.IDLETimeout = &to
+	if rt.IDLETimeout == nil {
+		if t := c.Duration("idle-timeout"); t > 0 {
```
treeder: Let's just make these cli.IntFlags, get rid of any duration stuff.

denismakogon: Done.

PR #17, opened 2017-08-30
add 'docker' runtime deploy test to cli/test.sh

amykang2020:
This adds 'docker' runtime 'fn deploy' test to the cli/test.sh, hence uses env DOCKER_USER like other tests in cli/test.sh


amykang2020:
'All checks have passed',  hence updated subject accordingly.

PR #19, opened 2017-08-31
Actually exits properly on deploy.

treeder:


deploy.go
```diff
@@ -95,7 +95,7 @@ func (p *deploycmd) scan(c *cli.Context) error {
 
 	err = filepath.Walk(wd, func(path string, info os.FileInfo, err error) error {
 		if path != wd && info.IsDir() {
-			return filepath.SkipDir
+			return nil
```
rdallman: this is the only thing i'm not sure about. this means where previously we would skip over directories that were not the directory we were looking for (if in fact, the path we were looking for ended up being a directory), now we will walk the contents of that directory. maybe this is what was broken (makes sense), fwiw it would still make sense to skip entire directory contents (save a lot of i/o) only in the case that the path we're seeking for and the current path do not override. e.g.

```
a/b
a/c

path=a/c/foo
```

now we will rummage through the contents of `a/b` even though we are certain that the file/dir we are looking for is not there. if somebody is doing this in their `Downloads` folder or something then it could be expensive.

deploy.go
```diff
@@ -106,18 +106,19 @@ func (p *deploycmd) scan(c *cli.Context) error {
...
-			fmt.Println(path, e)
+			// fmt.Println(path, e)
+			return fmt.Errorf("deploy error on %s: %v", path, err)
 		}
 
 		now := time.Now()
 		os.Chtimes(path, now, now)
 		walked = true
-		return e
+		return err
```
rdallman: if deploy returns a nil error (which we want) then this will walk the rest of the contents under `wd` (i assume this means working directory) even though we have already called `deploy`.

it would make sense to just use the `walked` value and use a sentinel error to return from `filepath.Walk` to indicate that we found the path and to not walk the rest of the directory's contents. again, i could imagine this taking a long time if a user does this in a directory like `$HOME` or in a really large repo.

rdallman:
sweet. seems like we should clean up how we're going about finding these things a little bit to improve ux, not sure if you want to do here or later, but left some notes. otherwise, it LGTM

deploy.go
rdallman: actually, due to this i wonder if it's possible that `deploy` gets called twice, on two different func files? eesh, need to stop that if that's true.

treeder: Yes, this is remnants of old crap, it looks for all funcfiles recursively and deploys them all. Thought of removing it since we only do singles for now, but I know I'll want it again when we have full app deploys.

rdallman: ahhh, makes sense.

PR #20, opened 2017-09-01
Required env vars.

### LINKED ISSUE #6: Support operating system variables inheriting while testing
So, when developer wants to test function, he uses `test.json` as a test contract
```
{
  "tests": [
    {
      "input": {
        "body": {
          "name": "Johnny"
        }
      },
      "output": {
        "body": {
          "message": "Hello Johnny"
        }
      }
    }
  ]
}
```
This proposal addressing certain cases when secrets needs to be ignored by VCS, for particular instance:
`i want to store my API keys away from Git but want to keep test.json in-tree`

So, proposing one for test contract attribute in addition that what already we have:
 - `from_os` or `from_env` (just need certain name that does not collide with `env` attribute of test.

@treeder @carimura @hibooboo2

treeder:
Enforces required env vars while testing. Will use defined env vars from local environment as well. 

Closes #6.

### DIFF
--- a/funcfile.go
+++ b/funcfile.go
@@ -27,15 +27,25 @@ type outputMap struct {
 }
 
 type fftest struct {
-	Name   string            `yaml:"name,omitempty" json:"name,omitempty"`
-	Input  *inputMap         `yaml:"input,omitempty" json:"input,omitempty"`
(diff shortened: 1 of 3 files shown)


treeder:
This is how it works for a user: https://github.com/fnproject/fn/pull/285/files#diff-e026c1b05ec782c09c02bbf147e5e811

PR #23, opened 2017-09-07
Changed fn deploy to use --app and fn init to use --name.

treeder:
FYI, `fn init` without `--name` defaults to directory name.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ vendor/
 _vendor*/
 /cli
 data/
+.vscode/
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ if you are using Node, put the code that you want to execute in the file `func.j
 Run:
 
 ```sh
-fn init [FUNCTION_NAME]
+fn init [--name FUNCTION_NAME]
 ```
 
 If you don't specify a name, the current directory name will be used.
@@ -33,7 +33,7 @@ If you don't specify a name, the current directory name will be used.
 If you want to override the convention with configuration, you can do that as well using:
 
 ```sh
-fn init [--runtime node] [--entrypoint "node hello.js"] [FUNCTION_NAME]
+fn init [--runtime node] [--entrypoint "node hello.js"] [--name FUNCTION_NAME]
 ```
 
 Or, if you want full control, just make a Dockerfile. If `init` finds a Dockerfile, it will use that instead of runtime and entrypoint.
@@ -205,7 +205,7 @@ path: /myfunc
 ### Creating a new function from source
 
 ```
-fn init hello --runtime ruby
+fn init --name hello --runtime ruby
 fn deploy --app myapp
 ```
 
--- a/README.md
+++ b/README.md
@@ -206,13 +206,13 @@ path: /myfunc
 
 ```
 fn init hello --runtime ruby
-fn deploy myapp /hello
+fn deploy --app myapp
 ```
 
 ### Updating function
 
 ```
-fn deploy myapp
+fn deploy --app myapp
 ```
 
 ### Testing function locally
--- a/deploy.go
+++ b/deploy.go
@@ -22,11 +22,10 @@ func deploy() cli.Command {
 	var flags []cli.Flag
 	flags = append(flags, cmd.flags()...)
 	return cli.Command{
-		Name:      "deploy",
-		ArgsUsage: "<appName>",
-		Usage:     "deploys a function to the functions server. (bumps, build, pushes and updates route)",
-		Flags:     flags,
-		Action:    cmd.scan,
+		Name:   "deploy",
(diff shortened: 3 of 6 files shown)


treeder:
Updating docs, don't merge.

rdallman:
`name` seems ambiguously, pardon me, named. our current verbiage has:

* app_name
* path
* route

is `name` not appropriately represented by any of the above 3? it seems like it's effectively route.name, but is that not 'path' now? is the intent here to remove 'path' from the vocabulary? if so, 'path' is unfortunately relatively entrenched atm.

rdallman:
oh, name is `image`, but without registry. mostly, the prior comment stands. the registry thing seems cumbersome if that's now a requirement (i.e. this whole thing does not seem trivial to get set up anymore)

treeder:
Not much has changed here except for adding `--app` and `--name` instead of guessing what the first param is (and neither are required). The Docker registry username thing was a must to change, didn't make sense in the first place. 

* `app_name` is the app name
* `name` is the function name
* `path` is the endpoint where the function is accessible in an app (usually == `name`, but can override here), also optional. 
* `route` put all the above together along with everything else in the func file and send it to your fn server and you have your route. User usually doesn't have to think about this though. We could change `routes` to something else, like `functions`?. But I think routes in the context of an app makes sense.

rdallman:
still trying to figure out exactly what 'name' is, i still think it should be a more descriptive word as 'name' can be just about anything.

> name is the function name

by this definition, it seems like it could be misinterpreted relatively easily as the path. but i take it now to mean that it's the name of an image that exists on some registry less the username for that image, where now we've chosen to redefine what 'registry' means, and 'registry' now means 'registry username' and not 'registry'. i'm not sure this vocabulary exists? not to be emphatic, but this seems really awful. FN_REGISTRY_USERNAME -- sure. making it required? still ick. now when i want to use http://my-frickin-registry.my-frickin-company.com what are we going to call that? FN_REGISTRY_ACTUAL_REGISTRY? overall i find the previously somehow agreed upon verbiage confusing as well as this proposed addition. `image` seems befitting, but now it's only the actual image name that's been namespaced already in your username and isn't what everybody in the world that has used docker hub actually expects an `image` name to be.

concretely i'd proposed using `image` instead of `name`, where `image` is an actual image name, such as `rdallman/yodawg`. this makes it obvious what's actually going on here, i don't see much point in masking the fact that an 'image' is a 'function'.

carimura:
i agree with the FN_REGISTRY comments... this occurred to me as well when setting that. FN_REGISTRY_USERNAME makes more sense, or FN_REGISTRY_USER

need to think on the name stuff some more. I didn't like my username in all the func.yaml's checked into repo, but I'm not sure how larger teams will work/share yet.

rdallman:
i've found the discussion of the mentioned topic on `FN_REGISTRY` -- i still find it confusing that we're assuming that registry is one of:

* a fully qualified 'registry' name, inclusive of a username
* potentially just a username for docker hub

images should be able to specify a registry on them and have this work (we are parsing for it, at least). i don't like that we have them separated and mandated now, personally, it seems like it should be entirely optional (and we can default to docker hub), and image names can stay image names (i.e. username+image name, e.g. `rdallman/yodawg`). this isn't necessarily my opinion this just seems to be how the container eco system at large seems to have structured things and i find it odd that we've decided to deviate (is there a good reason to do so?).

rdallman:
> I didn't like my username in all the func.yaml's checked into repo, but I'm not sure how larger teams will work/share yet.

yea, this is an interesting point. to me it just seems like we are optimizing for this case of 'hey i have this function and you should run it yourself, here's the code and a func.yaml, do `fn deploy` and run it yourself'. docker has done a nice job of having `Dockerfile` not contain any information about the name itself, to skirt around this issue, but the trade off is that every time you 'docker push' (which would equal a 'fn deploy' for us, in ideology) you __must__ specify the docker image name [and registry, if non-hub]. I think this could potentially make sense for us to require + dropping the image in func.yaml -- it would get rid of this issue, but we still need a good way to manage multiple functions on the whole (maybe this fits better into that).

treeder:
Can we please move this discussion to another issue?  This PR is a minor change to current behavior.

rdallman:
In part, sure. I still think `--name` should be `--image`, even if it's a little imperfect, and seems relevant to this discussion...

treeder:
All this does is change from first arg to a `--name` flag to set the `name` in the func.yaml.  This isn't meant to be a discussion about whether it should be called name or not, that's a much bigger discussion and a much bigger change.

rdallman:
true, that is the way things are.

rdallman:
circle failed. but seems fine to change implied positional arguments to be explicit flags now.

treeder:
Yep, fixing test.

PR #26, opened 2017-09-11
Skips name validation until deploy.

treeder:
Also fixes name check on patch route (so you can set config vars and what not).

### DIFF
--- a/deploy.go
+++ b/deploy.go
@@ -136,7 +136,17 @@ func (p *deploycmd) deploy(c *cli.Context) error {
 func (p *deploycmd) deployFunc(c *cli.Context, funcFilePath string) error {
 	funcFileName := path.Base(funcFilePath)
 
-	err := c.App.Command("bump").Run(c)
+	ff, err := loadFuncfile()
+	if err != nil {
+		return err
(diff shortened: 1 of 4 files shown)


carimura:
this worked for my vista demo setenv.sh... but can't comment on code itself. Also I'm trying to deal with other changed elements that are blocking my usual demo scripts (ie fn deploy now not working as before)

deploy.go
```diff
@@ -136,7 +136,17 @@ func (p *deploycmd) deploy(c *cli.Context) error {
 func (p *deploycmd) deployFunc(c *cli.Context, funcFilePath string) error {
 	funcFileName := path.Base(funcFilePath)
 
-	err := c.App.Command("bump").Run(c)
+	ff, err := loadFuncfile()
+	if err != nil {
+		return err
+	}
+
+	err = validateImageName(ff.ImageName())
```
rdallman: in the case it's not a --local we would be getting this from dockerpush already, this just guards against bumping and building i guess, and seems like we're loading the function file twice now just to get the error before bump. anyway, nit...

PR #33, opened 2017-09-15
Use an http query to find which version of the java runtime to use

hhexo:
Using version ranges in the pom means that maven will have to download _all_ poms to find which one is the latest. This is bad. However, we can't fix the version in the hardcoded boilerplate otherwise we'd have to change the hardcoded boilerplate on every build of the runtime.

We need a quick fix before we tackle the wider issue of having a boilerplate generation image which revs independently from the fn cli (basically https://github.com/fnproject/fn/issues/135 ), so for now we make an http query and parse the result to get a version number.

### DIFF
--- a/langs/java.go
+++ b/langs/java.go
@@ -9,6 +9,9 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"net/http"
+	"encoding/json"
+	"io"
 )
 
 // JavaLangHelper provides a set of helper methods for the lifecycle of Java Maven projects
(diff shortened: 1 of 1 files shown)


zootalures:
Note for readers - this is a (temporary?)  bugfix to prevent maven from spamming away resolving versions with maven version placeholders. 

Placeholders were put in to prevent us from having to rev the CLI every time we release. 

We'll review when we look at build helper containers - I think that will remove the need to make calls as the API version will be tagged within the build helper (or the helper can make the call)

PR #40, opened 2017-09-22
Bad regression fix and doesn't require node_modules dir anymore.

treeder:
Wasn't building on run anymore.


hibooboo2:
So this was removed and we're adding it back why was it removed in the first place?

nigeldeakin:
I have tested this and can confirm it makes the problem I reported go away (i.e. `fn run` will now do a build if needed)

denismakogon:
@hibooboo that one gone not so long ago, and we need to bring that back.

PR #46, opened 2017-09-25
adding Fn ascii to init command

carimura:


init.go
```diff
@@ -126,6 +126,15 @@ func (a *initFnCmd) init(c *cli.Context) error {
 		}
 	}
 
+	const runHeader = `
```
treeder: Seems a bit much for a command like this... 🤷‍♀️   Lots of extra lines with no real value.

carimura: it's either there or when you install the CLI... Serverless does it at their init command. I think it fits. People in theory won't be fn init'ing all the time.

carimura: Ping. Value is branding and small touches. Can we merge?

treeder:
Conflicts

carimura:
fixed

PR #47, opened 2017-09-26
Added ruby boilerplate and fixed fn test input bug.

treeder:


testfn.go
```diff
@@ -148,6 +148,11 @@ func (t *testcmd) test(c *cli.Context) error {
 
 func runlocaltest(target string, in *inputMap, expectedOut *outputMap, expectedErr *string, env map[string]string) error {
 	inBytes, _ := json.Marshal(in.Body)
+	if string(inBytes) == "\"\"" {
```
hibooboo2: Should be checking the error returned from marshal.

treeder: i'll add check for this, not sure why it's being skipped.

PR #48, opened 2017-09-26
Init, run and test use first arg as path now (like deploy).

treeder:
* `fn run myfunc` will run the function in the relative dir `myfunc`. 
* `fn init myfunc` will create a new dir and init the function in that dir. eg: `fn init --runtime ruby rubyfunc`

This also changes the behavior of `fn init`. It will not add the `name` value by default, instead it will use the directory name by default. This can be overridden by passing in `--name` flag or adding a `name` value to func.yaml.

### DIFF
--- a/init.go
+++ b/init.go
@@ -114,17 +114,21 @@ func initFn() cli.Command {
 func (a *initFnCmd) init(c *cli.Context) error {
 	wd := getWd()
 
+	var err error
 	path := c.Args().First()
 	if path != "" {
 		fmt.Printf("Creating function at: /%s\n", path)
 		dir := filepath.Join(wd, path)
 		// check if dir exists, if it does, then we can't create function
 		if exists(dir) {
-			return fmt.Errorf("directory %s already exists, cannot init function", dir)
-		}
-		err := os.MkdirAll(dir, 0755)
-		if err != nil {
-			return err
+			if !a.force {
+				return fmt.Errorf("directory %s already exists, cannot init function", dir)
+			}
+		} else {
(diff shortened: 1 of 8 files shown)


init.go
```diff
@@ -112,6 +112,26 @@ func initFn() cli.Command {
...
 func (a *initFnCmd) init(c *cli.Context) error {
+	wd := getWd()
+
+	path := c.Args().First()
+	if path != "" {
+		fmt.Printf("Creating function at: /%s\n", path)
+		dir := filepath.Join(wd, path)
+		// check if dir exists, if it does, then we can't create function
+		if exists(dir) {
+			return fmt.Errorf("directory %s already exists, cannot init function", dir)
```
denismakogon: Let users decide fail or not, if folder exist - allow user to override whole structure (`--force` maybe?)

treeder: Fixed.

hibooboo2:
So I read the code and still not entirely sure the purpose of this pr. When I checked it out and ran it it did not do what it sounded like you were describing. What should this pr change / make happen. When I run `fn init testfunc`

hibooboo2:
So it seems that you have to pass in runtime for init.. Or you have to have a folder called myfunc and that folder needs a standard func file eg func.go func.rb etc.. and you have to pass --force

- Have func file in folder called myfunc `fn init myfunc --force` Feels strange shouldnt need --force if folder is there but no func.yaml
- Have no folder call `fn init myfunc --runtime go` This works I like how it starts. Maybe have the ability to add --hot as well.

PR #55, opened 2017-09-29
Can set build_image and run_image

treeder:
If you want to use something other than our images. Or add packages to our images.


rdallman:
schwing

PR #57, opened 2017-09-29
Fix root function bug

### LINKED ISSUE #39: Conflicting function names in apps
From @tteggel: 

If I have 4 funcs

```
app.yaml
a/create/func.yaml
a/delete/func.yaml
b/create/func.yaml
b/delete/func.yaml
```

with no routes in the func.yaml then I get routes for /create and /delete pointing to b's images

treeder:
Merge #55 first, this is based off 55. 

Closes #39 here: https://github.com/fnproject/cli/pull/57/files#diff-deb23646b3f0c40871119853be51362aR193

### DIFF
--- a/testfn.go
+++ b/testfn.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


treeder:
Closes #39 here: https://github.com/fnproject/cli/pull/57/files#diff-deb23646b3f0c40871119853be51362aR193

testfn.go
```diff
@@ -65,6 +66,10 @@ func (t *testcmd) test(c *cli.Context) error {
 	if err != nil {
 		return err
 	}
+	// get name from directory if it's not defined
+	if ff.Name == "" {
+		ff.Name = filepath.Base(filepath.Dir(fpath)) // todo: should probably make a copy of ff before changing it
+	}
```
rdallman: seems like this `findAndParseFuncfile` and `if ff.Name == ""` is repeated all over, maybe we should look into doing this in front of the commands and handing over this info easier. not necessary here, just pointing this out...

PR #60, opened 2017-09-30
Will bump major or minor if user enters sq bracket bump major] or sq …

treeder:
…bracket bump minor]

eg:

"this is my commit [bump minor]" will bump the minor version, rather than patch (default).

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
-
 set -ex
 
 # ensure working dir is clean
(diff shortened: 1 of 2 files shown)


release.sh
```diff
@@ -19,7 +19,24 @@ if [ -z $(grep -m1 -Eo "[0-9]+\.[0-9]+\.[0-9]+" $version_file) ]; then
...
+bump_by="patch"
+last_commit=$(git log -1 --pretty=%B)
+x=$(docker run --rm -i treeder/strings contains "$last_commit" "[bump")
+if [ "$x" == "true" ]; then
+  x=$(docker run --rm -i treeder/strings contains "$last_commit" "[bump minor]")
+  if [ "$x" == "true" ]; then
+    echo "bumping minor version"
+    bump_by="minor"
+  fi
+  x=$(docker run --rm -i treeder/strings contains "$last_commit" "[bump major]")
```
rdallman: i don't think we need containers to run standard system utilities? seems a bit silly

release.sh
```diff
@@ -19,7 +19,24 @@ if [ -z $(grep -m1 -Eo "[0-9]+\.[0-9]+\.[0-9]+" $version_file) ]; then
...
+  if [ "$x" == "true" ]; then
+    echo "bumping major version"
+    bump_by="major"
+  fi
+fi
+# https://github.com/treeder/dockers/tree/master/bump
+docker run --rm -it -v $PWD:/app -w /app treeder/bump --filename $version_file $bump_by
```
rdallman: so can we just make this whole script the docker container? and then not shell out to other docker containers and just do the thing? we have this bump script littered across various places now and it's a bit cumbersome, we should just make a docker container that does the thing and run that in our ci.

treeder: Makes sense.

release.sh
treeder: If by silly you mean awesome. ;) The strings container will contain basically all the Go string function, so you can do a bunch of string ops easily.

treeder:
Going to wrap a bunch of stuff into one container like you said.

treeder:
Alright, check it now.

rdallman:
shwing

PR #65, opened 2017-10-08
Java pom.xml simplified -> build tag replaced with properties

AdamBien:
IMO the first pom.xml should be as simple as possible. Also: I would like to remove the dependencies to the API and tests (see e.g. https://www.youtube.com/watch?v=eWvj8ZVE5To, minute 1:46). 
The first project should be as lean as possible without any dependencies to the fn framework. 

Idea: e.g. fn init --withTests could generate integration tests as well.


hhexo:
It appears we had an issue with CircleCI builds for forks of the project... we just fixed it in our master, could you please rebase/merge? The checking build will then pass and we'll be able to merge.

Thanks again.

treeder:
@AdamBien sorry for this taking so long to review. Did you still want this fix in?  If so, please rebase and reopen.

PR #72, opened 2017-10-10
Add more docker image name validation rules based on specification, Fix #69

### LINKED ISSUE #69: 'fn deploy' failed if port number is specified in the docker registry URL
## To Reproduce the problem
1. Start docker registry with specified port number
  
  ```Shell
  $ docker run -d -p 5000:5000 --restart always --name registry registry:2
  ```

2. Execute `fn deploy` with registry option
  
  ```Shell
  $ fn deploy --app myapp --registry localhost:5000/mpppk
  Deploying fn-test to app: myapp at path: /fn-test
  Bumped to version 0.0.9
  Building image localhost:5000/mpppk/fn-test:0.0.9
  Sending build context to Docker daemon 17.41kB
  Step 1/8 : FROM funcy/go:dev as build-stage
   ---> 4cccab7fc828
  Step 2/8 : WORKDIR /function
   ---> Using cache
   ---> 6272ca9b2116
  Step 3/8 : ADD . /go/src/func/
   ---> fde238a0c90c
  Step 4/8 : RUN cd /go/src/func/ && go build -o func
   ---> Running in 7a49e0221081
   ---> 589dbab01ebf
  Removing intermediate container 7a49e0221081
  Step 5/8 : FROM funcy/go
   ---> 573e8a7edc05
  Step 6/8 : WORKDIR /function
   ---> Using cache
   ---> 59311302641b
  Step 7/8 : COPY --from=build-stage /go/src/func/func /function/
   ---> Using cache
   ---> 8314a0b63571
  Step 8/8 : ENTRYPOINT ./func
   ---> Using cache
   ---> ff1d0c216a7e
  Successfully built ff1d0c216a7e
  Successfully tagged localhost:5000/mpppk/fn-test:0.0.9
  ERROR: image name must have an owner and name, eg: username/myfunc. Be sure to set FN_REGISTRY env var or pass in --registry.
  ```

## Possible Fix
Fix `func validateImageName` in `common.go`.  
(I'll send a pull request soon.)

mpppk:
## Added rules
* Repository component can have DNS hostname prefix(Ref #69 )
* Tag values are limited to the set of characters `[a-zA-Z0-9_.-]`, except they may not start with a `.` or `-` character.
* Tags are limited to 128 characters
* A hostname component may not contain `_` characters
* Name components may contain lowercase characters, digits, and separators
* A separator is defined as a period, one or two underscores, or one or more dashes
* A name component may not start or end with a separator

## Ref
* https://github.com/moby/moby/blob/master/image/spec/v1.2.md


common.go
```diff
@@ -293,10 +295,36 @@ func validateImageName(n string) error {
...
+	}
+
 	if len(split2) < 2 {
 		return errors.New("image name must have an owner and name, eg: username/myfunc. Be sure to set FN_REGISTRY env var or pass in --registry.")
 	}
+
+	name := split2[len(split2)-1]
+	if !regexp.MustCompile(`^[a-z0-9]([a-z0-9_.-]*[a-z0-9])?$`).MatchString(name) ||
+		regexp.MustCompile(`\.{2}|_{3}`).MatchString(name) {
+		return errors.New("name only may contain lowercase characters, digits, and separators " +
```
AstromechZA: name only may -> name may only?

common_test.go
```diff
@@ -0,0 +1,41 @@
...
+		{"dockeruser/n:0.0.1", false},
+		{"dockeruser/na.me:0.0.1", false},
+		{"dockeruser/na__me:0.0.1", false},
+		{"localhost:5000/dockeruser/name:0.0.1", false},
+		{"dockeruser/name:" + longTagName, false},
+		{"dockeruser/name:" + tooLongTagName, true},
+		{"dockeruser/name:", true},
+		{"dockeruser/name:.0.0.1", true},
+		{"dockeruser/name:-0.0.1", true},
+		{"local_host:5000/dockeruser/name:0.0.1", true},
```
AstromechZA: can you add another test with an ipaddress:port as a test please, thanks

AstromechZA:
@mpppk have you signed the CLA yet? (https://github.com/fnproject/fn/blob/master/CONTRIBUTING.md)

mpppk:
@AstromechZA Today I signed CLA, thank you for your advice!

rdallman:
thanks for the great PR, this is really thorough. 

should we do this in the https://github.com/fnproject/fn server ? client side validation is nice as an added layer, but is brittle to maintain with server side validation checks (they can get out of sync / have bug mismatches). I have an inkling our server will probably gladly accept most of the things being gated by this patch, but I do think that the #69 case will 'just work' without `validateImageName`. 

I guess I propose we remove `validateImageName` from the CLI and let the server do validation. at present, it appears that the server does no validation of the image until we attempt to run it, at which point it may fail awkwardly since we are only parsing it and not checking for any errors. It would be nice if we did image validation at create/update time (not that the image exists, just that the image specified is valid) and returned this to the user at that point in time, from the server. What do you think?

treeder:
@mpppk any interest in moving this over to the server side?  Looks like a great addition.

mpppk:
@rdallman @treeder 
Sorry for late reply and thank you for your comments.
I'll send new pull request to [fnproject/fn](https://github.com/fnproject/fn) and close this PR. Thanks! :smile:

PR #73, opened 2017-10-11
Allow boilerplate generation to set build/run image for repeatable builds

hhexo:
Users have two competing requirements for their functions:

- They want builds to be repeatable (e.g. not be broken by a change in `latest`), and
- Different functions may require different versions of the language runtime

We have seen this several times with the Java FDK and runtime, where we even have an additional problem whereby our jars are also versioned in addition to the runtime image (and there's a correspondence between jars and runtime image). Cue the "don't merge anything during demo week!" requirement. 🙂 

Note that the main problem here is not only backwards compatibility (which we can strive to maintain): it is _forward_ compatibility. If I update my `fn` cli tool and do `fn init` and the generated boilerplate uses a feature that requires a later version of the runtime compared to the one I have, I cannot build/run my newly generated function!
I'd have to manually do a `docker pull` of the `latest` image runtime, but then my builds of my previous functions are not repeatable anymore.

Using `latest` is unsatisfactory. Therefore, users should lock their build and runtime images: since last week, this is possible using the `build_image` and `run_image` properties in the func.yaml. That's good. 👍 

In order to further improve usability, we think we should do this locking for the users if/when we generate the boilerplate for a new function: at the time, we can determine a version of dependencies (jars), build image, and run image and therefore we can lock them down.

However, this requires us to be able to modify the basic func.yaml at boilerplate generation time. In order to do so, this PR does the following:

- It moves the `funcfile` data structure into its own package, making it public
- It modifies the boilerplate generation contract so that `GenerateBoilerplate` can also modify properties of the `funcfile` that is going to be written.

This may not be the best way of refactoring the `funcfile` data structure - I tried to do the minimum movement of code that would satisfy the requirements. I am open to suggestions on how to do it better.
 
Still, we would really like to be able to modify the `funcfile` during boilerplate generation.

### DIFF
--- a/bump.go
+++ b/bump.go
@@ -78,7 +78,7 @@ func (b *bumpcmd) bump(c *cli.Context) error {
 	_, err = bumpItWd(wd, t)
 	return err
 }
-func bumpItWd(wd string, vtype VType) (*funcfile, error) {
+func bumpItWd(wd string, vtype VType) (*Funcfile, error) {
 	fn, err := findFuncfile(wd)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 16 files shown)


hhexo:
Closing because it's been done in a less invasive way in #108 .

PR #76, opened 2017-10-16
add back init route field mods

rdallman:
seems like at some point the code to mutate some of the default route fields
such as 'memory', 'format', 'timeout', 'idle_timeout' got removed, I'd guess
in the name of simplifying the func file output which I agree with (or perhaps a mistake?), but would
be nice to modify these fields too. we could omit them
if the defaults are captured, but that means we have to sync with the fn
server so not a huge fan of that (already an issue, really). anyway, the
defaults are back to being spat out, unless modified via flags.

found the ascii text pretty obnoxious in init, removed it. just created a
bunch of empty space in my shell that i had to scroll over, provides no value.
don't even like having any output when the command succeeds but that's too
neckbeard for this project probably...

now for tutorials we can do:

```
fn init --format http
```

instead of saying "open up your func.yaml in your favorite editor and add a
line...". also, nobody has to do that anymore, which is nice, too. only
works on select fields. i didn't want to do the whole update funcy dance
today, would be nice to embed a models.Route in the funcfile struct but it
doesn't have `yaml` tags and generates ugly fields :/ so, this works for now.

### DIFF
--- a/init.go
+++ b/init.go
@@ -99,7 +99,6 @@ func initFlags(a *initFnCmd) []cli.Flag {
 
 func initFn() cli.Command {
 	a := &initFnCmd{}
-	// funcfile := &funcfile{}
 
 	return cli.Command{
 		Name:        "init",
@@ -137,8 +136,9 @@ func (a *initFnCmd) init(c *cli.Context) error {
 		defer os.Chdir(wd) // todo: wrap this so we can log the error if changing back fails
 	}
 
-	rt := &models.Route{}
-	routeWithFlags(c, rt)
+	var rt models.Route
+	routeWithFlags(c, &rt)
+	a.bindRoute(&rt)
 
 	if !a.force {
 		_, ff, err := loadFuncfile()
@@ -150,17 +150,7 @@ func (a *initFnCmd) init(c *cli.Context) error {
 		}
 	}
 
-	const runHeader = `
-        ______
-       / ____/___
-      / /_  / __ \
-     / __/ / / / /
-    /_/   /_/ /_/`
-
-	fmt.Println(runHeader + "\n")
-
 	err = a.buildFuncFile(c)
-
 	if err != nil {
 		return err
 	}
@@ -174,8 +164,7 @@ func (a *initFnCmd) init(c *cli.Context) error {
 		}
 	}
 
-	ff := a.funcfile
-	if err := encodeFuncfileYAML("func.yaml", &ff); err != nil {
+	if err := encodeFuncfileYAML("func.yaml", &a.funcfile); err != nil {
 		return err
 	}
 	fmt.Println("func.yaml created.")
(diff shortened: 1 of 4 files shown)


rdallman:
totally made my examples in https://github.com/fnproject/fdk-go rely on this behavior already (no pressure, i'll forward all support requests ;)

treeder:
I really don't want to add the optional fields back in, it's minimal for a reason. Can you do this another way?

rdallman:
we could change the flags to default to "zero" instead of defaults so they won't be written to the file?

rdallman:
"type" is this way already

rdallman:
or i can just add the flags on the init command pointing to the funcfile fields instead

rdallman:
i removed the flag defaults [here](https://github.com/fnproject/cli/pull/76/commits/57ad82a093aac5476894f501f10da11b7beb9ac2) and now doing this:

```sh
cli init --format http --name rdallman/hot-gofdk
```

will make this:

```sh
✗: cat func.yaml
name: rdallman/hot-gofdk
version: 0.0.3
runtime: docker
format: http
```

if 'format' is omitted then it does not show up in the `func.yaml` either, as it was previously. 

should be alright now? i tested w/ creating a new route and updating one and worked as expected.

rdallman:
closes #80

init.go
```diff
@@ -150,17 +150,7 @@ func (a *initFnCmd) init(c *cli.Context) error {
 		}
 	}
 
-	const runHeader = `
-        ______
-       / ____/___
-      / /_  / __ \
```
carimura: @rdallman why remove this? want for fn init.

rdallman: see parent

rdallman: just obfuscates the 'your route has been updated' and usually when commands are loud i think something went wrong. this one is especially loud.

rdallman: ascii text in download command seems like what's really wanted. instead of bashing everyone's heads with 10 lines of ascii text every time they make a new function. i realize this is the Thiel way, but seems like a show once vs show all the time kind of thing. not much for fru fru clis myself though...

treeder: I'm with @rdallman (same comment I had when you added this in @carimura). Maybe in a less common command this would make sense, but it's annoying a fairly common one like init. +1 on install/download, much better place for it.

rdallman:
@carimura install does this now:

```
cli(init-format): curl -LSs https://raw.githubusercontent.com/fnproject/cli/init-format/install | sh
fn version 0.4.15

        ______
       / ____/___
      / /_  / __ \
     / __/ / / / /
    /_/   /_/ /_/`
```

(github doesn't format tabs, ignore spacing)

treeder:
@carimura needs your review.

init.go
carimura: okie dokie.

PR #79, opened 2017-10-18
Allow specifying custom port for fn start

irconan:
Adds a parameter to `fn start` to specify which port to bind on the host.

Defaults to the old hard-coded value of 8080

### DIFF
--- a/start.go
+++ b/start.go
@@ -25,6 +25,11 @@ func startCmd() cli.Command {
 				Name:  "detach, d",
 				Usage: "Run container in background.",
 			},
+			cli.IntFlag{
+				Name:  "port, p",
+				Value: 8080,
+				Usage: "Specify port number to bind to on the host.",
+			},
 		},
 	}
 }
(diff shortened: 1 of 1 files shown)


treeder:
Thank you!  LGTM, but waiting on CLA before we can merge.

irconan:
@treeder do you have details of the CLA process?

treeder:
There's this: http://www.oracle.com/technetwork/community/oca-486395.html which you can sign, scan and send back. 

But we're trying to make a fast and easy way to do it which will be integrated with GitHub.

carimura:
@irconan unfortunately our effort to move the CLA to digital signature is delayed so the links above are the best process at the moment. We'd love to have you as a contributor and we'll send a t-shirt once signed!

start.go
```diff
@@ -44,7 +49,7 @@ func start(c *cli.Context) error {
 		"--name", "fnserver",
 		"-v", fmt.Sprintf("%s/data:/app/data", wd),
 		"-v", "/var/run/docker.sock:/var/run/docker.sock",
-		"-p", "8080:8080",
+		"-p", fmt.Sprintf("%d:8080", c.Int("port")),
```
denismakogon: This is not enough, what if the user would use different service port? So, this port binding will make Fn server unreachable.

denismakogon: With recently merged code, you need to check env file for FN_PORT var and set exact service port to build the port binding, if FN_PORT is not mentioned - use default one.

treeder:
Closing for now, if you sign CLA, please reopen.

Issue #50, opened 2017-09-26
pagination of cli


rdallman:
need to figure out what api we want for this. we could bake into cli just getting all results or add pagination parameters to cli commands. in any case, we're going to start showing just the first 30 results here pretty soon in the cli, and we seem to be pushing the cli on people, so should probably have something for lack of ui / other tools.

treeder:
I at the very least (like you mentioned) have a --cursor flag, `fn x list --cursor c`. But could also stay open and page like `more`.  Or perhaps remember cursor in `~/.fn/somefile` then `fn x list --next`.

carimura:
doh. 23 days old. let's make a decision cause right now can only get first 30.

any input @hibooboo2 @denismakogon @derekschultz ?

denismakogon:
I'm on it. Swagger fixed, PRs for fn and fn_go are both ready. Once they would be merged, i'll jump on CLI.

hibooboo2:
Ref the prs to this issue Please.

denismakogon:
@hibooboo2 sure. There's one related - #77

hibooboo2:
Thats an issue. I meant your prs.

hibooboo2:
Ohh we do not have any pr's implementing this.
Nevermind

treeder:
Libs updated to 0.2.1

denismakogon:
well, it appeared that CLI still uses `funcy/functions_go` code, so it'll take some time to fix that in first place.

Issue #77, opened 2017-10-18
fn calls list ignores path filter


denismakogon:
Fn API allows to filter calls by app and route but CLI ignores path.
```
fn calls list --help
NAME:
   fn calls list - list all calls for specific app / route route is optional

USAGE:
   fn calls list <app> [route]
Deniss-MBP-2:cli denismakogon$ fn calls list application /env
ID: 01BWK116ZNY7WNT00000000000
App: application
Route: /traceback
Created At: 2017-10-16T19:42:23.605+03:00
Started At: 2017-10-16T19:42:23.605+03:00
Completed At: 2017-10-16T19:42:23.606+03:00
Status: success

ID: 01BWK115YHY7WNP00000000000
App: application
Route: /env
Created At: 2017-10-16T19:42:22.545+03:00
Started At: 2017-10-16T19:42:22.728+03:00
Completed At: 2017-10-16T19:42:23.587+03:00
Status: success
```

rdallman:
dead

PR #81, opened 2017-10-21
Pagination

denismakogon:
Adding pagination to calls API
```
fn calls list application --help
NAME:
   fn calls list - list all calls for the specific app. Route is optional

USAGE:
   fn calls list [command options] <app>

OPTIONS:
   --path value       function's path
   --cursor value     pagination cursor
   --from-time value  'start' timestamp
   --to-time value    'stop' timestamp
   --per-page value   number of calls to return (default: 30)
   
```

Closes: #50 #77

### DIFF
--- a/calls.go
+++ b/calls.go
@@ -56,8 +56,8 @@ func calls() cli.Command {
 					},
 					cli.Int64Flag{
 						Name:  "per-page",
-						Usage: "number of calls to return, if --per-page=-1 all calls will be listed",
-						Value: int64(-1),
+						Usage: "number of calls to return",
+						Value: int64(30),
 					},
 				},
 			},
--- a/calls.go
+++ b/calls.go
@@ -120,7 +120,7 @@ func (call *callsCmd) list(ctx *cli.Context) error {
 		if err != nil {
 			return err
 		}
-		res := fromTime_int64.Unix() / int64(time.Second)
+		res := fromTime_int64.Unix()
 		params.FromTime = &res
 
 	}
@@ -130,7 +130,7 @@ func (call *callsCmd) list(ctx *cli.Context) error {
 		if err != nil {
 			return err
 		}
-		res := toTime_int64.Unix() / int64(time.Second)
+		res := toTime_int64.Unix()
 		params.ToTime = &res
 	}
 	if ctx.Int64("per-page") > 0 {
--- a/calls.go
+++ b/calls.go
@@ -120,7 +120,7 @@ func (call *callsCmd) list(ctx *cli.Context) error {
 		if err != nil {
 			return err
 		}
-		res := fromTime_int64.UnixNano() / int64(time.Second)
+		res := fromTime_int64.Unix() / int64(time.Second)
 		params.FromTime = &res
 
 	}
@@ -130,7 +130,7 @@ func (call *callsCmd) list(ctx *cli.Context) error {
 		if err != nil {
 			return err
 		}
-		res := toTime_int64.UnixNano() / int64(time.Second)
+		res := toTime_int64.Unix() / int64(time.Second)
 		params.ToTime = &res
 	}
 	if ctx.Int64("per-page") != 0 {
--- a/calls.go
+++ b/calls.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/fnproject/cli/client"
 	fnclient "github.com/fnproject/fn_go/client"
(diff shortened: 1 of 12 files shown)


calls.go
```diff
@@ -82,14 +106,42 @@ func (call *callsCmd) list(ctx *cli.Context) error {
...
+		cursor := ctx.String("cursor")
+		params.Cursor = &cursor
 	}
+	if ctx.String("path") != "" {
+		route := ctx.String("path")
+		params.Path = &route
+	}
+	if ctx.String("from-time") != "" {
+		fromTime := ctx.String("from-time")
+		fromTime_int64, err := time.Parse(time.RFC3339, fromTime)
```
rdallman: `date --rfc-3339=date` will generate these, but I believe we need the nanosecond version and users may trip up on this. taking `unix` time here (as the api does) is just an int and may be less error prone.  thoughts ?

calls.go
```diff
@@ -82,14 +106,42 @@ func (call *callsCmd) list(ctx *cli.Context) error {
...
+		if err != nil {
+			return err
+		}
+		res := fromTime_int64.UnixNano() / int64(time.Second)
```
rdallman: can just do `Unix()` here, no need to take nanos and divide seconds out.

calls.go
denismakogon: each, `Unix` would be fine, I thought that nanoseconds would work, but they don't, we expect timestamps to be an int64 representation of timestamp in seconds.

anyways, PR updated.

calls.go
denismakogon: addressed

calls.go
```diff
@@ -33,9 +34,32 @@ func calls() cli.Command {
...
+						Usage: "'start' timestamp",
+					},
+					cli.StringFlag{
+						Name:  "to-time",
+						Usage: "'stop' timestamp",
+					},
+					cli.Int64Flag{
+						Name:  "per-page",
+						Usage: "number of calls to return",
+						Value: int64(30),
```
rdallman: just omit this imo, the server will sanely default to 30 if <= 0 is provided and we can change that in the future and have it fall out here.

denismakogon: There's a small problem, `cli.Int64Flag` will set the default value to 0 and will show it to users, so, keeping this default will at least be good for UX. That's why I don't want to move it away.

calls.go
```diff
@@ -82,14 +106,42 @@ func (call *callsCmd) list(ctx *cli.Context) error {
...
+		route := ctx.String("path")
+		params.Path = &route
+	}
+	if ctx.String("from-time") != "" {
+		fromTime := ctx.String("from-time")
+		fromTime_int64, err := time.Parse(time.RFC3339, fromTime)
+		if err != nil {
+			return err
+		}
+		res := fromTime_int64.Unix() / int64(time.Second)
```
rdallman: i think we can remove all the `/ int64(time.Second)`, the api for `Unix()` already returns exactly what we want https://golang.org/pkg/time/#Time.Unix and doing the division will end up changing the time to a different time: https://play.golang.org/p/t6dmjESpV2

I'm still wary of RFC3339 vs unix seconds in api (not just for reasons like this). generating unix time in shell is really easy: `date +%s` and also easy to use `bc` to offset it by 1 hour or whatever, and I'm not sure how flexible the RFC3339 date parser is in go.

calls.go
```diff
@@ -82,14 +106,42 @@ func (call *callsCmd) list(ctx *cli.Context) error {
...
+		params.FromTime = &res
+
+	}
+	if ctx.String("to-time") != "" {
+		toTime := ctx.String("to-time")
+		toTime_int64, err := time.Parse(time.RFC3339, toTime)
+		if err != nil {
+			return err
+		}
+		res := toTime_int64.Unix() / int64(time.Second)
```
rdallman: this one too

client/api.go
```diff
@@ -16,26 +14,38 @@ const (
 )
 
 func Host() string {
+	h, err := HostURL()
+	if err != nil {
+		panic(err)
```
rdallman: would prefer to output this on `os.Stderr` and then `os.Exit(1)` (preferably from bubbling the error up) -- users are specifying the `API_URL` directly and we should expect it to be wrong, `panic` makes more sense in cases where something shouldn't happen.

calls.go
rdallman: makes sense

deploy.go
```diff
@@ -274,9 +274,6 @@ func setRootFuncInfo(ff *funcfile, appName string) {
 
 func (p *deploycmd) updateRoute(c *cli.Context, appName string, ff *funcfile) error {
 	fmt.Printf("Updating route %s using image %s...\n", ff.Path, ff.ImageName())
-	if err := resetBasePath(p.Configuration); err != nil {
```
rdallman: not sure what this was? double checking it's ok?

rdallman: makes sense i think, doing this at client level now

denismakogon: yeah, it's on the responsibility of a client, this method was quite confusing and its purpose what not clear, it was just appending the value of API_URL and `v1`, so fn_go does this every time we're hitting any of its API.

version.go
```diff
@@ -1,47 +1,35 @@
...
+	// version is also there, but it shouldn't
+	// dropping base path to get appropriate URL for request eventually
+	t.BasePath = ""
+	version_client := fnclient.New(t, reg)
+	v, err := version_client.GetVersion(nil)
 	if err != nil {
 		return err
 	}
-	fmt.Println("Server version", v.Version)
+	fmt.Println("Client version: ", Version)
```
rdallman: this will mean that if one can't communicate with the server they can't see the client version. for downloads and in general for users to check their cli version, it would be nice to not have to have a live fn. at a minimum, can we move this back before the check of the server version?

I would suggest instead of returning the error to possibly put a `?` in place of the server version here, vs. showing the user:

```sh
$ fn --version
Client version: 0.4.2
HTTP client timeout: unable to reach http://thefntoruleallfn.com:8080/version
exit code 1
```

muting the error is kind of meh, but it's the version endpoint, and they'll find out soon enough if they try to do anything else. would be nice imo to see:

```sh
$ fn --version
Client version: 0.4.2
Server version: ?
```

denismakogon: addressed

denismakogon:
@treeder can you merge?

PR #92, opened 2017-11-16
Add a --selinux flag to `fn start` to deal with SELinux permissions

### LINKED ISSUE #91: Deal with SELinux permissions when starting the Fn container in `fn start`
On systems where the SELinux security policies are enforced, `fn start` is not able to mount the current directory as a writable volume in the spawned Fn Docker container. This causes Fn to be unable to write to the default database.

Of course it is possible to use a different db (e.g. mysql). Another way to solve this is for the user to tell SELinux to allow writes to the current directory, with `chcon -R -t svirt_sandbox_file_t <path>`. 

However, we can do better.
Docker supports a `:z` option for volume mounts, which will do the SELinux magic to set the proper permissions upon mounting the volume. If we change the cli to provide that option when it starts up the container, it will be transparent to the user and `fn start` will "just work" like it should.

Interesting reference articles about this:

- https://www.projectatomic.io/blog/2015/06/using-volumes-with-docker-can-cause-problems-with-selinux/
- https://www.projectatomic.io/blog/2016/03/dwalsh_selinux_containers/

hhexo:
Closes: #91 

Using the `:Z` option on the docker volume mounts when starting the Fn container will cause SELinux to grant read/write permission to them _specifically to the Fn container_.
This is the safest way of dealing with SELinux permissions as it does not require to generally lower the `svirt_sandbox_file_t` security setting.

### DIFF
--- a/start.go
+++ b/start.go
@@ -25,6 +25,10 @@ func startCmd() cli.Command {
 				Name:  "detach, d",
 				Usage: "Run container in background.",
 			},
+			cli.BoolFlag{
+				Name:  "selinux",
+				Usage: "Run container with the right privileges on a SELinux system.",
+			},
 		},
 	}
 }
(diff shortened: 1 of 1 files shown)


hhexo:
Wait... this makes it access the database fine, but it still doesn't run containers through the /var/docker.sock mount...

hhexo:
Since running fn locally needs access to the system's docker socket, and that access means the container can run arbitrary processes, we might as well run the container privileged.

I will add a `--privileged` flag to `fn start` so that a user on SELinux knows what they're doing when running fn locally.

hhexo:
Actually even this flag is not enough - now the containers can run, but processes inside containers cannot access filesystems in the containers themselves. It's a bit puzzling.

hhexo:
I'm closing this while I research a proper solution.

PR #94, opened 2017-11-20
cli: introducing docker mode

skinowski:
Preferred way to start fn is using docker in docker. However,
this is not available in Windows. With this change, adding
an option to specify this where by default preferred method
is docker in docker if applicable.

### DIFF
--- a/start.go
+++ b/start.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
+	"runtime"
 	"syscall"
 
 	"github.com/urfave/cli"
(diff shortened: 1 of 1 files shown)


start.go
```diff
@@ -34,18 +40,33 @@ func start(c *cli.Context) error {
 	if c.String("log-level") != "" {
 		denvs = append(denvs, "GIN_MODE="+c.String("log-level"))
 	}
-	// Socket mount: docker run --rm -it --name functions -v ${PWD}/data:/app/data -v /var/run/docker.sock:/var/run/docker.sock -p 8080:8080 fnproject/functions
-	// OR dind: docker run --rm -it --name functions -v ${PWD}/data:/app/data --privileged -p 8080:8080 fnproject/functions
+
+	// docker-in-docker or socker bind mount?
+	isDind := true
+	if c.String("docker-mode") == "docker-in-docker" && runtime.GOOS == "windows" {
```
denismakogon: this is not only windows-specific, on Mac i don't really rely on dind mode rather than running alpine image with fn binary inside and have my standalone docker.

treeder:
Can we roll this back?  `fn start` isn't intended to use docker-in-docker because it makes a bad development experience. The CLI most likely won't be used on servers, it's almost purely for development.

PR #100, opened 2017-11-23
Build and release ARM-compatible binary

denismakogon:
Closes: https://github.com/fnproject/fn/issues/431


Makefile
```diff
@@ -17,12 +17,18 @@ dep-up:
...
+	GOOS=darwin go build -o fn
+	GOOS=linux go build -o fn
+	go build -o fn
 	./test.sh
 
 release:
 	GOOS=linux go build -o fn_linux
 	GOOS=darwin go build -o fn_mac
 	GOOS=windows go build -o fn.exe
+	GOOS=linux GOARCH=amd64 go build -o fn_linux_arm64
```
treeder: Isn't this just the normal arch?  Did you mean to use `arm`?

denismakogon: Yeah. Fixing.

denismakogon:
@treeder can we merge this one? Also solves https://github.com/fnproject/fn/issues/431

treeder:
@denismakogon want to rebase this?

denismakogon:
yeah, would do, thanks

denismakogon:
@treeder can you review this one?

Makefile
```diff
@@ -16,13 +16,20 @@ dep:
 dep-up:
 	dep ensure
 
-test: build
+
+test:
+	GOOS=windows go build -o fn
```
treeder: Why this change?  Doesn't seem to make much sense.

denismakogon: because we don't know if we can compile fn until that moment when CI attempts to release fn

denismakogon: I think that is the valid reason to check if fn compiles for various platforms we consider as "supported"

treeder: I just mean why build five times, one overwriting the next. Seems like the old way was fine.

denismakogon: that's the point, each next step overwrites previous and the last step will produce linux-compatible binary that CI can use for further testing.

treeder: I don't think this is necessary and it doesn't build the correct one for whoever is running this command.

denismakogon: i extracted that to its own namespace:` test-cross-compile`

rikgibson:
Given no movement here in 7 months I'd suggest closing this.

3...2...1...

rdallman:
think there's a lot to do to support arm https://github.com/fnproject/fn/issues/1128 e.g. `fn start` won't work either even with this. would be cool eventually, just needs some <3

rikgibson:
Think this should just be closed then - it can always be reopened if ever Arm support becomes a priority.

Anprotaku:
Is there any update on this?

PR #101, opened 2017-11-27
Fn-prefix things

denismakogon:


client/api.go
```diff
@@ -24,7 +24,7 @@ func Host() string {
 }
 
 func HostURL() (*url.URL, error) {
-	apiURL := os.Getenv("API_URL")
+	apiURL := os.Getenv("FN_API_URL")
 	if apiURL == "" {
```
carimura: the only thing that's confusing is for those that have the old API_URL format set, it'll now ignore it and set it to localhost:8080... potentially confusing people. Maybe we can put in a deprecated message for those that have API_URL set?

carimura: I'm going to merge this for now and commit a possible fix

PR #102, opened 2017-11-30
checks for API_URL, not critical but seems nice for people who alread…

carimura:
Errors out if user has API_URL set.

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -1,13 +1,13 @@
 package client
 
 import (
-	"os"
+	"fmt"
 	"net/url"
+	"os"
 
 	fnclient "github.com/fnproject/fn_go/client"
 	openapi "github.com/go-openapi/runtime/client"
 	"github.com/go-openapi/strfmt"
-	"fmt"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


client/api.go
```diff
@@ -25,7 +25,12 @@ func Host() string {
 
 func HostURL() (*url.URL, error) {
 	apiURL := os.Getenv("FN_API_URL")
+
 	if apiURL == "" {
+		if os.Getenv("API_URL") != "" {
+			fmt.Fprint(os.Stderr, "Error: API_URL is deprecated, please use FN_API_URL.")
```
denismakogon: this might be just a warning, don't have to be so critical

carimura:
I tried println. Wasn't pretty. How? :)

On Nov 30, 2017 01:23, "Denis Makogon" <notifications@github.com> wrote:

*@denismakogon* commented on this pull request.
------------------------------

In client/api.go
<https://github.com/fnproject/cli/pull/102#discussion_r154021144>:

>  	if apiURL == "" {
+		if os.Getenv("API_URL") != "" {
+			fmt.Fprint(os.Stderr, "Error: API_URL is deprecated, please use
FN_API_URL.")

this might be just a warning, don't have to be so critical

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/fnproject/cli/pull/102#pullrequestreview-80118253>, or mute
the thread

client/api.go
denismakogon: I mean that suppose to be enough to write the message to STDERR like you did, I'm not sure that we should exit in case if FN_API_URL is not set.

carimura: since it initializes on every command, it prints the warning like 20 times. I can fix by introducing a global bool but feels unnecessary .... this is all temporary until all current users switch to FN_API_URL and then we can just delete these lines.

PR #108, opened 2017-12-05
Optionally fixate build and runtime images in func.yaml on init 

zootalures:
Reason for PR: 

Fn builds are generally unpredictable as they depend on :latest images for build and run in most cases. 

If a language binding supports working out specific tags for build and run  then let it set that in build_image and runtime_image on fn init.

### DIFF
--- a/common.go
+++ b/common.go
@@ -202,7 +202,10 @@ func writeTmpDockerfile(helper langs.LangHelper, dir string, ff *funcfile) (stri
 	dfLines := []string{}
 	bi := ff.BuildImage
 	if bi == "" {
-		bi = helper.BuildFromImage()
+		bi, err = helper.BuildFromImage()
+		if err != nil {
+			return "", err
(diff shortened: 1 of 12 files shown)


shaunsmith:
I don't see any docs updated to describe the changes in PR #108?  What are the semantics and how is it used? It looks like FN_JAVA_FDK_VERSION can be used for configuration but I grepped for references and didn’t find any other than in java.go.

zootalures:
Yep we've had this behaviour for a few months (it just works a bit diferently now)  but you are right tht docs are missing - any thoughts  is a the right place to doc this?  (it's a CLI feature but specific to java?)

zootalures:
Thinking about it  - if the other FDKs were added on FN init (they need to be) then this is a generic param

shaunsmith:
True, FN_FDK_VERSION could be enough?

PR #113, opened 2017-12-09
--runtime flag output runtime's name

stevensu1977:
let's --runtime is friendly useful
use runtime name replace ext file name .

### DIFF
--- a/init.go
+++ b/init.go
@@ -49,7 +49,7 @@ var (
 )
 
 func init() {
-	for rt := range fileExtToRuntime {
+	for _, rt := range fileExtToRuntime {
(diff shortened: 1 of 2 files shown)


denismakogon:
There's one thing that has to be fixed as well, https://github.com/fnproject/cli/blob/6627438abb21eccf1e534e0f4b477fac09ffb046/init.go#L36, for python specifically, there are 3 runtimes: python, python2.7, python3.6

stevensu1977:
I change python boilerplate, just use print("hello,world"), it's can be deploy python/python3.6 runtime

treeder:
@stevensu1977 can you sign CLA so we can look at this? Reopen if you want to continue this.

Issue #110, opened 2017-12-06
Add error to call CLI API


denismakogon:
CLI needs newer fn_go and its features like `call.error` thing (indicates why the particular call failed).

Issue #109, opened 2017-12-06
Better pagination for calls


denismakogon:
At this moment CLI doesn't do that at its best but should be.

PR #115, opened 2017-12-11
Better pagination

denismakogon:
Closes: https://github.com/fnproject/fn/issues/560
Closes: https://github.com/fnproject/cli/issues/110
Closes: https://github.com/fnproject/cli/issues/109


calls.go
```diff
@@ -138,15 +141,24 @@ func (call *callsCmd) list(ctx *cli.Context) error {
...
+				return errors.New(e.Payload.Error.Message)
+			default:
+				return err
+			}
 		}
+
+		printCalls(resp.Payload.Calls)
+
+		if resp.Payload.NextCursor == "" {
+			break
```
rdallman: don't think this is really such a good idea. when we have real users, they are going to have thousands, tens of thousands, maybe millions of calls and this will make __a ton__ of api calls, take a _really_ long time for the CLI command to return and is very likely not what the user wants. we should allow a user to be able to get this data (and we do, through the API), but we absolutely should not make it the default behavior [or even make it all that easy]. it seems reasonable as a default to return the last hour of calls, but not every single call.

rdallman: even last hour is a lot, really. but i think from the CLI using time ranges makes sense. most of the time, at least, when i'm searching calls i want ones from last few hours (from some test), not some 'last N' calls (which is harder to gauge).

denismakogon: Okay, PR updated, i tend to agree that last hour calls may work but with pretty intensive Fn usage it may be hit the same problem, but i do believe that users will use pagination opts for that.

rdallman: thanks. we can adjust, some bound is better than none. for me, usually looking at last few calls is all i want, and if i want to do something with a lot of data i end up using the api client in go to write a little program, but that's me. 

i'm skeptical of apps / routes listing all by default too. it defeats having pagination to begin with (same with calls). i'd be slightly more comfortable with truncating it to 100 or 1000 and forcing a user to paginate, b/c users can do dumb things sometimes (make 1M routes...). one of those things where i don't think it would be the end of the world to truncate and output that the user should use the client bindings / api to sift through all the data and not the cli, cli pagination by user just seems gnarly and to be discouraged.

treeder: I think any listing call should just return a fixed number (ie: per-page). Agree with Reed that it definitely should not auto paginate through all results, there's almost no scenario where that would be a good idea. We could use the 'per-page' flag though to page through multiple result sets though. eg: if user passed in `--per-page 1000`, then loop until 1000 results.

start.go
```diff
@@ -29,7 +29,7 @@ func startCmd() cli.Command {
 	}
 }
 
-func start(c *cli.Context) error {
```
treeder: Why did you remove this error response?

common.go
```diff
@@ -178,7 +178,7 @@ func dockerVersionCheck() error {
 	}
 	vMin, err := semver.NewVersion(minRequiredDockerVersion)
 	if err != nil {
-		return fmt.Errorf("our bad, sorry... please make an issue.", err)
+		return fmt.Errorf("our bad, sorry... please make an issue, original error: %v", err)
```
treeder: Can you remove "original error", it's not a different error, it's the only error.

denismakogon:
Rebased. @treeder @rdallman can you revisit your reviews to make this land to the master branch?

treeder:
Did you fix the possible infinite loop?  I still wonder why this doesn't just call the API and return whatever it gets. If the API has a 30 default, use that. If the API has a max of 100 (or whatever it is), let the API decide and return whatever it gives you, instead of looping over API calls here.

denismakogon:
@treeder @rdallman can you revisit this one?

rdallman:
it shouldn't infinite loop since it's checking against `resp.Payload.Cursor == ""`

what we have is okay. i guess it makes sense to change to use as default `--n 100` instead of `--from-time=123` so that if a user makes a lot of calls (in last hour) they don't get thousands of calls back by default. there is probably some interplay between these. i.e. if `--from-time` is specified and `-n` isn't, we could choose not to limit the results to `-n`. i'm game for changing to use `-n` primarily, defaulting to not setting `-n` and taking the server default first page, but also allowing use of `--from-time` and `--to-time` as added already.  @treeder ?

treeder:
I think we should just default to nothing and let the API be the defaults. I think that is what user's would expect and we won't run into the issue of doing 2 calls to get the results (eg: if server defaulted to less than 100). 

Otherwise, seems fine by me.

calls.go
```diff
@@ -138,15 +145,27 @@ func (call *callsCmd) list(ctx *cli.Context) error {
...
 
-	resp, err := call.client.Call.GetAppsAppCalls(&params)
-	if err != nil {
-		switch e := err.(type) {
-		case *apicall.GetAppsAppCallsNotFound:
-			return errors.New(e.Payload.Error.Message)
-		default:
-			return err
+	var resCalls []*models.Call
+	for {
```
rdallman: we need to get rid of this too to only return 100 unless they specify the last hour (instead of defaulting to it)

denismakogon: I see this a bit differently. `--per-page` is useless that's why it does make sense to replace one with `--n` which is defaulted to 100 objects, so the API will return its default number of calls.
If `--from-time` not specified CLI will return 100 calls from the last hour.

rdallman: sorry, I should have commented on the default last hour line.

denismakogon: i see, fixed.

apps.go
```diff
@@ -86,42 +86,65 @@ func apps() cli.Command {
...
+				// this is the one getting called, not sure what the one above is?
+				return fmt.Errorf("%v", e.Payload.Error.Message)
+			default:
+				return fmt.Errorf("%v", err)
+			}
 		}
+
+		resApps = append(resApps, resp.Payload.Apps...)
+
+		howManyMore := c.Int64("per-page") - int64(len(resApps)+len(resp.Payload.Apps))
```
rdallman: seems like this should be `-n` like calls, too?

calls.go
```diff
@@ -55,9 +55,9 @@ func calls() cli.Command {
 						Usage: "'stop' timestamp",
 					},
 					cli.Int64Flag{
-						Name:  "per-page",
+						Name:  "n",
 						Usage: "number of calls to return",
-						Value: int64(30),
+						Value: int64(100),
```
rdallman: 0 and use API default like apps?

routes.go
```diff
@@ -88,6 +88,17 @@ func routes() cli.Command {
 				Usage:     "list routes for `app`",
 				ArgsUsage: "<app>",
 				Action:    r.list,
+				Flags: []cli.Flag{
+					cli.StringFlag{
+						Name:  "cursor",
+						Usage: "pagination cursor",
+					},
+					cli.Int64Flag{
+						Name:  "per-page",
```
rdallman: `-n` too ?

start.go
```diff
@@ -82,10 +82,8 @@ func start(c *cli.Context) error {
 		case err := <-done:
 			if err != nil {
 				log.Println("error: processed finished with error", err)
+				return err
```
rdallman: if the container exits cleanly this will just hang (why, is less clear), I think we should move this back

rdallman:
i think just need to tighten up the `-n` and `-per-page` disparity between calls and apps and routes (and the default value to just use api value). it's kind of weird to have `-n` paired with `cursor` since cursors are not necessarily just a resource id/name it's not clear what to put in them in conjunction with `-n` but anyway we can leave it hanging off on the side, `-n` everywhere seems fine at least.

routes.go
denismakogon: yes, adjusting.

calls.go
denismakogon: this is no longer `--per-page`, so the API would return default number of objects per page, so, i think this should remain as is (defaulted to `100`)

rdallman:
sweet, thanks

denismakogon:
thanks

Issue #5, opened 2017-08-25
"build:" output elided


jan-g:
The following func.yaml seems to have broken somewhere between 0.3.63 and 0.3.69:

```
name: fn-test/test-4
version: 0.0.1
runtime: java

path: /test
timeout: 450

build:
  - mvn package dependency:copy-dependencies -DincludeScope=runtime -DskipTests=true -Dmdep.prependGroupId=true -DoutputDirectory=target

cmd: com.fnproject.fn.integration.ExerciseEverything::handleRequest
```

specifically, the "build" commands are not run prior to packaging the content with a supplied Dockerfile. Is this accidental? This was pretty useful behaviour for runtimes like the Java one where it's often the case that a function may be a module of a larger project.

jan-g:
(Specifically, 0.3.63 works; this is broken in 0.3.64 and onwards.)

jan-g:
Actually - it's working, it's just not showing any output.

I'm fine(ish)* with this in the success case, but it's eating any diagnostics on failure.

* package managers and corporate proxies and hanging downloads are a thing that I'd rather the user got to see rather than sitting for n minutes with no feedback.

treeder:
Is this still an issue?

zootalures:
I don't think this should be swallowed on a verbose build

PR #117, opened 2017-12-12
Hides docker build unless error, shows dots instead.

treeder:
dec-11-2017 16-30-53

Or when error occurs:
dec-11-2017 16-39-52

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ release:
 	GOOS=linux go build -o fn_linux
 	GOOS=darwin go build -o fn_mac
 	GOOS=windows go build -o fn.exe
-	docker run --rm -v ${PWD}:/go/src/github.com/fnproject/cli -w /go/src/github.com/fnproject/cli golang:alpine go build -o fn_alpine
+	# Uses fnproject/go:x.x-dev because golang:alpine has this issue: https://github.com/docker-library/golang/issues/155 and this https://github.com/docker-library/golang/issues/153
+	docker run --rm -v ${PWD}:/go/src/github.com/fnproject/cli -w /go/src/github.com/fnproject/cli fnproject/go:1.9-dev go build -o fn_alpine
 
 .PHONY: install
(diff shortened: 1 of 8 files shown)


rdallman:
fwiw i never minded the docker output even for succeeding builds, at a minimum because sometimes i'm on a not so hot network and the build step is stuck yanking down some gigantic base image to use or doing some gigantic apt-get install and i'm just curious what it's stuck on. with this, in that case since it's buffering the output until there's an error, i'll never know where it got stuck when i give up after 20 minutes, throw my computer across the room, and mumble vulgar words about whales. not sure of the benefits here other than simple noise reduction and possibly hiding that we use docker, but idk, pros > cons to me I think to just show it. `fn deploy` is noisy, yea, but when that output is useful, it's useful, and ctrl+k makes the bad man go away easily enough.

carimura:
@rdallman agreed, think we could simply add a -v for verbose mode for those that want to peek behind curtain.

treeder:
Mostly just because the docker build output is so lengthy. And it's a little bit startling on first go as it  makes up 99% of all the output. For example:

Old:
screen shot 2017-12-11 at 11 44 03 pm

New:
screen shot 2017-12-11 at 11 44 15 pm

treeder:
Added a global `--log-level` flag for any command to use, but any command that builds the container will not support this. See images in previous comment for example using this flag.

treeder:
Also, if there's an error it will show this:

screen shot 2017-12-11 at 11 47 06 pm

treeder:
BTW, could have verbose too (which would just be log-level=debug), just using log-level to be consistent with other things. Also more flexible since someone may want log-level fatal or something (ie: no output).

rdallman:
linking for relevance https://github.com/fnproject/cli/issues/5 (oldie but goodie)

rdallman:
verbose mode seems like a fine compromise. fwiw heroku's cli is super duper noisy by default and i think most people find it a nice experience

treeder:
@rdallman --log-level work for you?

rdallman:
seems more like a server-side config, also have to type `-log-level=debug` which isn't perfectly intuitive. `-v` and `-vv` [and even `-vvv`] are pretty standard CLI syntax for verbose mode.

treeder:
Ok, will add -v

treeder:
Changed to --verbose and -v.

NOTE: this overrides the `-v` for version now.

Issue #104, opened 2017-11-30
Supported runtimes in fn help init are out of date


tteggel:
Actual:
```
fn help init
...
   --runtime value           choose an existing runtime - .go, .rb, .php, .rs, .js, .py, .cs, .fs, .java
...
```
Expected:
List of runtimes from https://github.com/fnproject/cli/blob/master/langs/base.go#L17

denismakogon:
Would be nice to have CLI command to list runtimes

treeder:
should get the help output from the list in base.go.

treeder:
Fixed.

Issue #78, opened 2017-10-18
fn --help USAGE section contains broken link


r4zzz4k:
```sh
> fn --help
fn 0.4.11

Fn command line tool

USAGE:
   Check the docs at https://github.com/fnproject/fn/blob/master/fn/README.md

ENVIRONMENT VARIABLES:
...
```
The link to [README.md](https://github.com/fnproject/fn/blob/master/fn/README.md) is broken.
Of course the CLI itself provides enough information to actually get started anyway, so it's probably just a cosmetic issue.

treeder:
This has been fixed.

Issue #64, opened 2017-10-07
Should fn init --runtime values be values rather than file extensions?


DazWilkin:
https://github.com/fnproject/cli/blob/9352fa3c55028c5c0cd1d33181cd5e3b0630f461/init.go#L52

```
fn init --help
...
choose an existing runtime - .js, .cs, .java, .go, .rb, .py, .php, .rs, .fs
...
```

Golang (`.go` vs. `go`) does not provide a good example but...

```
fn init --runtime=python

        ______
       / ____/___
      / /_  / __ \
     / __/ / / / /
    /_/   /_/ /_/

Runtime: python
func.yaml created.
```

I think this *may* be addressable with flipping `rt` from key to value:

```
for _,rt := range fileExtToRuntime {
	fnInitRuntimes = append(fnInitRuntimes, rt)
}
```

So:
```
strings.Join(fnInitRuntimes, ", ")
```
Produces:
```
rust, dotnet, java, node, python, php, go, ruby, dotnet
```

And perhaps (once supported), dedup'ing C#, F# etc. to just one `dotnet`

treeder:
Fixed.

PR #118, opened 2017-12-12
Lang helpers define what runtime's they can handle.

### LINKED ISSUE #112: --runtime help information confused
--runtime help list file extend name like ".go",".java", is not runtime correct name,
other thing help information have no version information , like "python2.7", "python3.6", "node" ,"lambda-node-4" , it's not friendly useful .

may be we can use registry model, let  every XXXLangHelper reg self with runtime name

treeder:
This helps with various issues/discrepancies so we don't have to maintain things in various places. Now the lang helpers decide what runtime strings they can handle. 

This is a somewhat annoying change as it adds a few extra methods to LangHelper which is already getting a bit out of control. Should think of a nicer way to do LangHelpers at some point. 

Closes #112, #64, #104, #78

### DIFF
--- a/init.go
+++ b/init.go
@@ -223,7 +223,7 @@ func (a *initFnCmd) buildFuncFile(c *cli.Context) error {
 		helper = langs.GetLangHelper(a.Runtime)
 	}
 	if helper == nil {
-		fmt.Printf("Init does not support the %s runtime, you'll have to create your own Dockerfile for this function", a.Runtime)
+		fmt.Printf("Init does not support the %s runtime, you'll have to create your own Dockerfile for this function.\n", a.Runtime)
 	} else {
 		if a.Entrypoint == "" {
 			a.Entrypoint, err = helper.Entrypoint()
--- a/init.go
+++ b/init.go
@@ -32,28 +32,6 @@ import (
 	"github.com/urfave/cli"
 )
 
-var (
-	fileExtToRuntime = map[string]string{
-		".go":   "go",
-		".js":   "node",
-		".rb":   "ruby",
-		".py":   "python",
-		".php":  "php",
-		".rs":   "rust",
-		".cs":   "dotnet",
-		".fs":   "dotnet",
-		".java": "java",
-	}
-
-	fnInitRuntimes []string
-)
-
-func init() {
-	for rt := range fileExtToRuntime {
-		fnInitRuntimes = append(fnInitRuntimes, rt)
-	}
-}
-
 type initFnCmd struct {
 	force bool
 	funcfile
@@ -73,7 +51,7 @@ func initFlags(a *initFnCmd) []cli.Flag {
 		},
 		cli.StringFlag{
 			Name:        "runtime",
-			Usage:       "choose an existing runtime - " + strings.Join(fnInitRuntimes, ", "),
+			Usage:       "choose an existing runtime - " + langsList(),
 			Destination: &a.Runtime,
 		},
 		cli.StringFlag{
@@ -95,6 +73,18 @@ func initFlags(a *initFnCmd) []cli.Flag {
 	}
 
 	return append(fgs, routeFlags...)
+}
+
+func langsList() string {
+	// var b bytes.Buffer
+	// for _, h := range helpers {
+	// 	b.Write([]byte(h.))
+	// }
+	allLangs := []string{}
+	for _, h := range langs.Helpers() {
+		allLangs = append(allLangs, h.LangStrings()...)
+	}
+	return strings.Join(allLangs, ", ")
 }
 
 func initFn() cli.Command {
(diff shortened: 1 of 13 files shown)


init.go
```diff
@@ -97,6 +75,18 @@ func initFlags(a *initFnCmd) []cli.Flag {
 	return append(fgs, routeFlags...)
 }
 
+func langsList() string {
+	// var b bytes.Buffer
```
zootalures: ditch commentcode

init.go
```diff
@@ -225,18 +215,17 @@ func (a *initFnCmd) buildFuncFile(c *cli.Context) error {
...
-		a.Runtime = rt
-		fmt.Printf("Found %v function, assuming %v runtime.\n", rt, rt)
+		fmt.Printf("Found %v function, assuming %v runtime.\n", helper.Runtime(), helper.Runtime())
 	} else {
 		fmt.Println("Runtime:", a.Runtime)
+		helper = langs.GetLangHelper(a.Runtime)
 	}
-	helper := langs.GetLangHelper(a.Runtime)
 	if helper == nil {
 		fmt.Printf("Init does not support the %s runtime, you'll have to create your own Dockerfile for this function", a.Runtime)
```
zootalures: Not direclty related to this change but setting an unknown runtime gives a garbled error due to missing newline here: 
```
OCliffe@OCLIFFE-mac /t/foojazz> fn init --runtime asdas
Runtime: asdas
Init does not support the asdas runtime, you'll have to create your own Dockerfile for this functionERROR: could not detect entrypoint or cmd for asdas, use --entrypoint and/or --cmd to set them explicitly
```

langs/base.go
```diff
@@ -6,6 +6,32 @@ import (
...
+func registerHelper(h LangHelper) {
+	helpers = append(helpers, h)
+}
+
+func Helpers() []LangHelper {
+	return helpers
+}
+
 //used to indicate the default supported version of java
 const defaultJavaSupportedVersion = "9"
```
zootalures: no longer needed

langs/lambda_node.go
```diff
@@ -4,6 +4,20 @@ type LambdaNodeHelper struct {
...
+}
+func (h *LambdaNodeHelper) Runtime() string {
+	return h.LangStrings()[0]
+}
+
+func (lh *LambdaNodeHelper) LangStrings() []string {
+	return []string{"lambda-nodejs4.3", "lambda-node-4"}
+}
+func (lh *LambdaNodeHelper) Extensions() []string {
+	return []string{".py"}
```
zootalures: shouldn't this be .js?

langs/python.go
```diff
@@ -10,6 +10,24 @@ type PythonLangHelper struct {
...
 
+func (h *PythonLangHelper) Handles(lang string) bool {
+	return defaultHandles(h, lang)
+}
+func (h *PythonLangHelper) Runtime() string {
+	return h.LangStrings()[0]
+}
+
+// TODO: I feel like this whole versioning thing here could be done better. eg: `runtime: python:2` where we have a single lang, but support versions in tags (like docker tags).
+func (lh *PythonLangHelper) LangStrings() []string {
```
zootalures: This feels a bit messy (as h.Version is already specified on construction) - right now  I prefer if we set the runtime name (for java and python)  (and maybe aliases) on struct init ? thoughts?

treeder: Agree, feels very messy (see my comment). 

I think we should have tags for versions rather than explicit runtimes with the versions built in. eg: `runtime: java:9` . runtime is java, version is 9. This would apply to all languages for versioning. 

I didn't attempt fixing that issue in this PR though. It uses a similar (bad) scheme as before: https://github.com/fnproject/cli/pull/118/files#diff-a3639713bb27239dd39d24f7d2fc9678L25, but still fixes the discrepancy issues. 

Made a follow up #119 to do the version thing properly.

init.go
treeder: fixed.

langs/base.go
treeder: fixed

langs/lambda_node.go
treeder: fixed

langs/python.go
zootalures: ack, thanks

init.go
```diff
@@ -97,6 +75,14 @@ func initFlags(a *initFnCmd) []cli.Flag {
 	return append(fgs, routeFlags...)
 }
 
+func langsList() string {
+	allLangs := []string{}
+	for _, h := range langs.Helpers() {
+		allLangs = append(allLangs, h.LangStrings()...)
+	}
```
denismakogon: given loop can be a bit simplified somehow in terms of the number of lists it creates

langs/lambda_node.go
```diff
@@ -4,6 +4,22 @@ type LambdaNodeHelper struct {
...
 
+func (h *LambdaNodeHelper) Handles(lang string) bool {
+	return defaultHandles(h, lang)
+}
+func (h *LambdaNodeHelper) Runtime() string {
+	return h.LangStrings()[0]
+}
+
+func (lh *LambdaNodeHelper) LangStrings() []string {
+	return []string{"lambda-nodejs4.3", "lambda-node-4"}
```
denismakogon: isn't this a good time to add version tag to lambda-node as well?

init.go
treeder: example?

langs/lambda_node.go
treeder: I think lambda helper will go away soon, instead support it in FDK. But regardless, I didn't do anything with versioning here.

PR #120, opened 2017-12-14
Initial Kotlin support + add 'deploy' in the 'help' list of commands

delabassee:
'deploy' is simply missing from the help

### DIFF
--- a/main.go
+++ b/main.go
@@ -69,6 +69,7 @@ LEARN MORE:
 		lambda(),
 		version(),
 		calls(),
+		deploy(),
 		logs(),
 		testfn(),
 		buildServer(),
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -69,6 +69,7 @@ LEARN MORE:
 		lambda(),
 		version(),
 		calls(),
+		deploy(),
```
treeder: This is already there: https://github.com/fnproject/cli/pull/120/files#diff-7ddfb3e035b42cd70649cc33393fe32cR77

treeder: Oh, I see, it's not showing --help.

treeder:
Should probably mention that this PR also adds Kotlin support. ;)

delabassee:
PR title fixed!
That's what happens when you submit a PR during a conf.call :expressionless:

Issue #122, opened 2017-12-14
fn start doesn't support fnserver configuration options


denismakogon:
We'd like to have DB, MQ, LogStore, etc. env vars configurable during `fn start`. Since Fn CLI uses exec, it's worth adding env file which is simple and requires less code to be changed.

rdallman:
yea, this might be a good idea. on the other hand, i do think we were actively trying to discourage using `fn start` for any kind of complex deploy / testing (there are obvious issues with logs, signal handling, proc handling)

denismakogon:
I do agree, that start cmd is just suppose to be simple thing to start with, so having an ability to set for instance datastore that is different from SQLite or different mq bring less complexity to run fn server. But, yeah, it’s must stay simple and not be something that meant to be for production, just local thing.

PR #123, opened 2017-12-14
Use docker env file as source of Fn server configuration opts

denismakogon:
Closes: #122

### DIFF
--- a/start.go
+++ b/start.go
@@ -17,6 +17,10 @@ func startCmd() cli.Command {
 		Usage:  "start a functions server",
 		Action: start,
 		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name:  "log-level",
+				Usage: "--log-level DEBUG to enable debugging",
+			},
 			cli.BoolFlag{
 				Name:  "detach, d",
 				Usage: "Run container in background.",
@@ -40,6 +44,9 @@ func start(c *cli.Context) error {
 		"-v", "/var/run/docker.sock:/var/run/docker.sock",
 		"--privileged",
 		"-p", "8080:8080",
+	}
+	if c.String("log-level") != "" {
+		args = append(args, "-e", fmt.Sprintf("GIN_MODE=%v", c.String("log-level")))
 	}
 	if c.String("config") != ""{
 		args = append(args, "--env-file", c.String("config"))
--- a/start.go
+++ b/start.go
@@ -26,8 +26,8 @@ func startCmd() cli.Command {
 				Usage: "Run container in background.",
 			},
 			cli.StringFlag{
-				Name: "config, c",
-				Usage: "Absolute path to Fn server configuration options file.",
+				Name: "env-file",
+				Usage: "Path to Fn server configuration file.",
 			},
 		},
 	}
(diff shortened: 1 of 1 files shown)


start.go
```diff
@@ -17,25 +17,19 @@ func startCmd() cli.Command {
 		Usage:  "start a functions server",
 		Action: start,
 		Flags: []cli.Flag{
-			cli.StringFlag{
```
treeder: Why did you remove this?

start.go
```diff
@@ -44,10 +38,10 @@ func start(c *cli.Context) error {
 		"--name", "fnserver",
 		"-v", fmt.Sprintf("%s/data:/app/data", wd),
 		"-v", "/var/run/docker.sock:/var/run/docker.sock",
-		"-p", "8080:8080",
```
treeder: Why change this?

start.go
denismakogon: because it's not necessary anymore, you can put one to env file or just set FN_LOG_LEVEL to DEBUG and you'll be fine.

start.go
denismakogon: because it can be really whatever it can, you have FN_PORT which would change port, so it's not always 8080.

start.go
treeder: How would FN_LOG_LEVEL make it into the container?

start.go
treeder: I'd much prefer this: https://github.com/fnproject/cli/pull/79 vs what you have here.

start.go
denismakogon: through env file, that's the purpose of this whole patch - let user pass fn server opts to the container.

treeder: Ok, well please don't make .env file required, you're making a lot of assumptions here.

start.go
```diff
@@ -17,25 +17,19 @@ func startCmd() cli.Command {
...
-				Name:  "log-level",
-				Usage: "--log-level DEBUG to enable debugging",
-			},
 			cli.BoolFlag{
 				Name:  "detach, d",
 				Usage: "Run container in background.",
 			},
+			cli.StringFlag{
+				Name: "config, c",
+				Usage: "Absolute path to Fn server configuration options file.",
```
treeder: And why not call it `--env-file` like docker? Also, does it really require an absolute path?

start.go
denismakogon: it's not required by default

start.go
denismakogon: okay, will revert.

start.go
treeder: I'm saying, please keep log-level there. I don't want to have to make an env file to get debugging output.

start.go
denismakogon: because we're not trying to reimplement docker CLI here, moreover, `--env-file` name is not verbose or helpful. The name of the opt should be very clear, that's why it was named as `--config`

denismakogon: well, path doesn't suppose to be absolute, but it's more clear than even docker CLI helper string: `Read in a file of environment variables`

treeder: But this isn't really a config file, it's a file containing env vars. If we ever support a config file, it probably (almost surely) wouldn't be in this format. This file format dates back to the Ruby days: https://github.com/bkeepers/dotenv

Also, the the flags should match docker run (like -d) since this is just a convenience command that passes right through to docker run.

denismakogon:
@treeder so, can we merge this?

treeder:
If you change it to --env-file to match docker, then yes.

denismakogon:
@treeder done.

PR #125, opened 2017-12-18
Fixes regression generating func.yaml. auto detected runtimes weren't…

treeder:
… added to file.

Closes https://github.com/fnproject/fn/issues/608

### DIFF
--- a/init.go
+++ b/init.go
@@ -24,7 +24,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-
 	"strings"
 
 	"github.com/fnproject/cli/langs"
@@ -34,7 +33,7 @@ import (
 
 type initFnCmd struct {
 	force bool
-	funcfile
+	ff    *funcfile
 }
 
 func initFlags(a *initFnCmd) []cli.Flag {
@@ -42,7 +41,7 @@ func initFlags(a *initFnCmd) []cli.Flag {
 		cli.StringFlag{
 			Name:        "name",
 			Usage:       "name of the function. Defaults to directory name.",
-			Destination: &a.Name,
+			Destination: &a.ff.Name,
 		},
 		cli.BoolFlag{
 			Name:        "force",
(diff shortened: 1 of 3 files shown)


init.go
```diff
@@ -34,15 +33,15 @@ import (
 
 type initFnCmd struct {
 	force bool
-	funcfile
+	ff    *funcfile
```
treeder: Made this explicit, was hard to follow what was going on previously.

init.go
```diff
@@ -136,29 +135,31 @@ func (a *initFnCmd) init(c *cli.Context) error {
...
+	err = a.buildFuncFile(c) // TODO: Return LangHelper here, then don't need to refind the helper in generateBoilerplate() below
 	if err != nil {
 		return err
 	}
 
-	runtimeSpecified := a.Runtime != ""
+	runtimeSpecified := a.ff.Runtime != ""
 
-	if runtimeSpecified && a.Runtime != funcfileDockerRuntime {
+	// TODO: why don't we treat "docker" runtime as just another language helper? Then can get rid of several Docker
```
denismakogon: good point, been using `runtime: docker` for binaries, so it's really a nice-to-have thing.

init_test.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		t.Fatalf("ERROR: Failed run `init` command: err: %v", err)
+	}
+
+	ffname := "func.yaml"
+	b, err := ioutil.ReadFile(ffname)
+	if err != nil {
+		t.Fatalf("could not open %s for parsing. Error: %v", ffname, err)
+	}
+	fmt.Printf("FUNC: %+v\n", string(b))
```
denismakogon: `t.Logf` is better

init_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/urfave/cli"
+	yaml "gopkg.in/yaml.v2"
+)
+
+func TestInit(t *testing.T) {
+
+	testname := "test-init"
+	testdir, err := ioutil.TempDir("", testname)
+	if err != nil {
+		t.Fatalf("ERROR: Failed to make tmp test directory: err: %v", err)
```
denismakogon: better to use `err.String()`, not sure that `%v` would give human-readable err representation.

treeder: It does.

treeder:
Removed println's

PR #128, opened 2017-12-23
Ship logs as text

denismakogon:
Follow-up of https://github.com/fnproject/fn/pull/615

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -40,8 +40,8 @@
 [[projects]]
   name = "github.com/aws/aws-sdk-go"
   packages = ["aws","aws/awserr","aws/awsutil","aws/client","aws/client/metadata","aws/corehandlers","aws/credentials","aws/credentials/ec2rolecreds","aws/credentials/endpointcreds","aws/credentials/stscreds","aws/defaults","aws/ec2metadata","aws/endpoints","aws/request","aws/session","aws/signer/v4","internal/shareddefaults","private/protocol","private/protocol/json/jsonutil","private/protocol/jsonrpc","private/protocol/query","private/protocol/query/queryutil","private/protocol/rest","private/protocol/restjson","private/protocol/xml/xmlutil","service/lambda","service/sts"]
-  revision = "71472b3d628115a0650f4851e48dde483cc84776"
-  version = "v1.12.45"
+  revision = "82ad808f2307df0776c038bfd7ea85440a35c02e"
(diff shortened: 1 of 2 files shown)


denismakogon:
@treeder @rdallman

rdallman:
@denismakogon this hard breaks anyone with an out of date fn server, not sure if we want to add code to handle a json blob in case we get one or just let it break and force updates (we'll get bugs filed)

treeder:
@denismakogon where does this ask for text/plain? How is it not getting the default json?

denismakogon:
@treeder
https://github.com/fnproject/fn_go/blob/master/client/operations/operations_client.go#L42

treeder:
I still feel like this is going to cause problems down the road when we want to add more fields to this response, but we will never be able to use them because now we've committed to only getting text back. Just as this change is backwards incompatible, it will also be when we realize we want it reverted.

rdallman:
we're still maintaining a json return for that endpoint that we can change our clients to use again, if that ever happens. for logs, specifically, i just don't see it happening, and even when/if it's ever a problem we can add another endpoint with that stuff; it seems more likely we add fields to the get-call-by-id endpoint. 

this change was backwards compatible, if done properly. that's the catch.

rdallman:
in the cli, we just need to handle the returned content type being json or plain text and act accordingly. problem solved.

rdallman:
well, in all clients, really

denismakogon:
As long as we use auto-generated clients I don’t see any simple way to do so, swagger doesn’t look kinda friendly for keeping backward compatible API.

rdallman:
https://github.com/OAI/OpenAPI-Specification/issues/270#issuecomment-284062885 seems to imply swagger 3 has the capability, haven't investigated fully yet

rdallman:
https://swagger.io/docs/specification/media-types/ seems like we can add one for `application/json` and one for `text/plain` -- gonna try it out

rdallman:
@denismakogon so, i think maybe for swagger let's keep the old json for now to not break people (revert changes in fn to swagger), revert this patch to use json again, change swagger from 2 to 3 (see: https://blog.readme.io/an-example-filled-guide-to-swagger-3-2/ ) and then we can have multiple media types as linked above and move swagger over. interested in doing this?

rdallman:
we can keep the plain text server code, mostly just need to revert clients to use same json response as they were.

Issue #82, opened 2017-10-25
RFE: No directory should be created when 'fn init' fails


delabassee:
`fn init xxx --runtime yyy` will fails because the 'yyy' runtime doesn't exist but still, an empty 'xxx' dir will be created.

AstromechZA:
I'd like to pick this one up.

Reproduced on current master below:

```
$ ~/projects/go_workspace/src/github.com/fnproject/cli/fn init xxx --runtime yyy
Creating function at: /xxx
Runtime: yyy
Init does not support the yyy runtime, you'll have to create your own Dockerfile for this function.
ERROR: could not detect entrypoint or cmd for yyy, use --entrypoint and/or --cmd to set them explicitly
$ ls
xxx/
```

AstromechZA:
Fix just needs some review at https://github.com/fnproject/cli/pull/129

AstromechZA:
Fix is merged. Just need someone to close this issue since the PR didn't do it.

PR #129, opened 2017-12-28
Added runtime specification whitelist check to top of init cli.

AstromechZA:
Fixes issue #82.

Before:

```
$ fn init xxx --runtime yyy
Creating function at: /xxx
Runtime: yyy
Init does not support the yyy runtime, you'll have to create your own Dockerfile for this function.
ERROR: could not detect entrypoint or cmd for yyy, use --entrypoint and/or --cmd to set them explicitly
$ ls
xxx/
```

After:

```
$ fn init xxx --runtime yyy
ERROR: Init does not support the 'yyy' runtime.
$ ls
```

### DIFF
--- a/init.go
+++ b/init.go
@@ -98,6 +98,18 @@ func initFn() cli.Command {
 func (a *initFnCmd) init(c *cli.Context) error {
 	wd := getWd()
 
+	var rt models.Route
+	routeWithFlags(c, &rt)
+	a.bindRoute(&rt)
+
+	runtimeSpecified := a.ff.Runtime != ""
+	if runtimeSpecified {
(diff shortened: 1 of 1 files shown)


AstromechZA:
I'm unsure why its saying I haven't signed the CLA. I do appear on the signatories list as "Benjamin Meier - Fn Project - astromechza". I've double-checked that I used the same email address on the form and the git log (and github).

treeder:
@AstromechZA I'll look into it and ensure you get setup properly.

carimura:
@AstromechZA this is on our end.. fixing the bot. Check should clear up soon (next hour or so). Thanks for the PR!

PR #135, opened 2018-01-04
cli: add FN_DEADLINE if Timeout is specified

skinowski:
Fixes #133

Issue #130, opened 2017-12-29
Revert function's version if deploy fails


denismakogon:
At this moment every time, we call a deploy cmd (`fn deploy`) it bumps function's version, but the new version of a function should appear only in case if deploy was successful, in other cases that should remain the same, so we need to revert function's version in case if deploy fails.

hibooboo2:
Is this not addressed with: #136 ?

denismakogon:
Yes, that PR fixed that problem.

hibooboo2:
Maybe add the closes for this issue to that pr then?

Issue #133, opened 2018-01-02
Broken "test" command


denismakogon:
`fn test` doesn't set `FN_DEADLINE` which is required by FDKs (Go, Python).

treeder:
When was FN_DEADLINE added and what is it for?  I don't see it documented anywhere.

denismakogon:
FN_DEADLINE defines when function will timeout, each FDK implements deadline-based requests handler, so each function that uses FDK will fail without FN_DEADLINE.

Well, it’s not documented because developers don’t have an access to that.

rdallman:
@denismakogon they do have access to it, we should doc it. fdk's just handle it nicely

denismakogon:
Well, if they use FDK they have nothing to do with that because it’s already applied to their code. Indeed, worth documenting.

skinowski:
I'll document this.

Issue #132, opened 2018-01-02
Broken "run" command


denismakogon:
1. `fn run` doesn't set FN_DEADLINE.
2. `fn run` is not aware of json format.

```
echo 'data' | fn run --format json
Building image denismakogon/test:0.0.3 
ERROR: Format json not supported
```
3. `fn run` doesn't read format from func file

zootalures:
@denismakogon I think we fixed this, could you verify

PR #136, opened 2018-01-04
Fixes problems with: "deploy", "run" and "test"

denismakogon:
Closes #135 #132 #133 #130

### DIFF
--- a/deploy.go
+++ b/deploy.go
@@ -225,14 +225,22 @@ func (p *deploycmd) deployFunc(c *cli.Context, appName, baseDir, funcfilePath st
 	}
 	fmt.Printf("Deploying %s to app: %s at path: %s\n", funcfile.Name, appName, funcfile.Path)
 
-	funcfile, err := bumpVersion(funcfile, Patch)
+	oldVersion := funcfile.Version
+	newFuncfile, err := bumpIt(funcfilePath, Patch)
 	if err != nil {
 		return err
 	}
 	// TODO: this whole funcfile handling needs some love, way too confusing. Only bump makes permanent changes to it.
 
-	_, err = buildfunc(c, funcfilePath, funcfile, p.noCache)
+	_, err = buildfunc(c, funcfilePath, newFuncfile, p.noCache)
 	if err != nil {
+		// if build was faulty we need to rollback funcfile
+		// version because broken build doesn't mean new version of the function
+		funcfile.Version = oldVersion
+		storeErr := storeFuncfile(funcfilePath, funcfile)
+		if storeErr != nil {
+			return storeErr
+		}
 		return err
 	}
 
@@ -240,11 +248,6 @@ func (p *deploycmd) deployFunc(c *cli.Context, appName, baseDir, funcfilePath st
 		if err := dockerPush(funcfile); err != nil {
 			return err
 		}
-	}
-
-	err = storeFuncfile(funcfilePath, funcfile)
-	if err != nil {
-		return err
 	}
 
 	return p.updateRoute(c, appName, funcfile)
--- a/deploy.go
+++ b/deploy.go
@@ -225,11 +225,10 @@ func (p *deploycmd) deployFunc(c *cli.Context, appName, baseDir, funcfilePath st
 	}
 	fmt.Printf("Deploying %s to app: %s at path: %s\n", funcfile.Name, appName, funcfile.Path)
 
-	funcfile2, err := bumpIt(funcfilePath, Patch)
+	funcfile, err := bumpVersion(funcfile, Patch)
 	if err != nil {
 		return err
 	}
-	funcfile.Version = funcfile2.Version
 	// TODO: this whole funcfile handling needs some love, way too confusing. Only bump makes permanent changes to it.
 
 	_, err = buildfunc(c, funcfilePath, funcfile, p.noCache)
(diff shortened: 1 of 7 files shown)


run.go
```diff
@@ -201,37 +211,65 @@ func runff(ff *funcfile, stdin io.Reader, stdout, stderr io.Writer, method strin
...
+	h := http.Header{}
+	h.Set("FN_CALL_ID", "12345678901234567890123456")
+	h.Set("FN_METHOD", method)
+	h.Set("FN_REQUEST_URL", LocalTestURL)
+	h.Set("FN_APP_NAME", "myapp")
+	h.Set("FN_PATH", ff.Path)
+	h.Set("FN_FORMAT", ff.Format)
+	h.Set("FN_MEMORY", strconv.Itoa(int(ff.Memory)))
+	h.Set("FN_TYPE", "sync")
+	h.Set("FN_DEADLINE", strfmt.DateTime(
```
skinowski: this should use the real ff.Timeout if defined.

denismakogon: indeed, if not defined we should use default value which is 30s.

denismakogon:
@treeder can you review, please?

funcfile.go
```diff
@@ -8,7 +8,7 @@ import (
 	"path/filepath"
 	"strings"
 
-	yaml "gopkg.in/yaml.v2"
```
treeder: Can you keep these explicit ones in here, I don't know what it is, but my tools (`goimports` maybe?) often rewrite these to the github URL.

test.sh
```diff
@@ -8,7 +9,7 @@ if [[ -z "$FN_REGISTRY" ]]; then
 fi
 $fn --version
 
-go test $(go list ./... | grep -v /vendor/ | grep -v /tests)
+go test -v $(go list ./... | grep -v /vendor/ | grep -v /tests)
```
treeder: Can you remove verbose?  Don't need to see everything unless it fails.

funcfile.go
denismakogon: this particular import works fine without any alias, so, why it is needed?

denismakogon: anyways, reverted.

deploy.go
```diff
@@ -225,11 +225,10 @@ func (p *deploycmd) deployFunc(c *cli.Context, appName, baseDir, funcfilePath st
 	}
 	fmt.Printf("Deploying %s to app: %s at path: %s\n", funcfile.Name, appName, funcfile.Path)
 
-	funcfile2, err := bumpIt(funcfilePath, Patch)
+	funcfile, err := bumpVersion(funcfile, Patch)
```
hibooboo2: I believe that this change will have unintended consequences.. I am not sure did you look at the side effects of doing this?

deploy.go
```diff
@@ -243,6 +242,11 @@ func (p *deploycmd) deployFunc(c *cli.Context, appName, baseDir, funcfilePath st
 		}
 	}
 
+	err = storeFuncfile(funcfilePath, funcfile)
```
hibooboo2: Should we be modifying the func file on deploy? Will this cause unintended changes?

denismakogon: At this moment, we modify func file every time we do deploy because we bump version no matter of if deploy was successful or not. So the behaviour should be different, we define new version, run deploy with that, if the deploy was successful (images was built correctly, and pushed as well) we update funcfile struct with a new version and then write it back to func file.

That is the only flow i see here which is correct. Despite that, i tested that on the different type of use cases like the bad registry, bad build, good build with the bad registry, good build with the good registry.

deploy.go
denismakogon: there are no unintended consequences because bumpVersion only bumps the version in `funcfile` struct, it doesn't write to the file, but `bumpIt` does both which is wrong, CLI should do the following steps:
 1. bump version at `funcfile` struct only
 2. run deploy
 3. write file

So, in case of failure at step 2 func version will remain the same which is good, failed build must not change the function version.

treeder: What happens if two succeeds, but three doesn't, that would be bad. Might be better to write file before deploying.

denismakogon: That's exactly what it is right now, and this is what it makes function version grow with every unsuccessful deploy which is the wrong behaviour. If step 3 is not successful it means that someone (not the CLI) corrupted func file, at that moment we it's not our fault and we should not do something unexpected.

treeder: The problem is that if we deploy without being 100% sure the func file is updated, then the next time you deploy, you will basically overwrite the previous version, which is bad. It's better to have your version numbers grow since that has zero chance of any side effects.

denismakogon: “Next time deploy” really means that the previous deploy for 100% unsuccessful, so, obviously, broken deploy doesn’t mean new version in any possible case. So, if we try to rerun the deploy it would override built image only in one case - someone corrupted func.yml which means that function version (from func.yml) corrupted as well, so we don’t know which previous versions function had. Override is an expected behaviors. I’d like us to follow exactly the same logic that docker uses - it doesn’t declare new tag unless that was explicitly specified, so the CLI should do the same - build-bump, but not bump-no-matter-what-then-build.

hibooboo2:
closes #130

denismakogon:
Closes: #130

hibooboo2:
Need to add that in the first comment or it will not close.

deploy.go
treeder: You understand the risk right?  

Why does it matter if the bump happens before the build? Because a number is skipped?  Who cares? There's zero risk that way.

denismakogon: Okay, here's how i see it. We do what we were doing before (update func file with new version before starting the build), but we should definitely revert func file in case if build was faulty.

denismakogon:
Closing in favor of #174

PR #140, opened 2018-01-07
Fixed envvar name for fn start and tweak to help text

### LINKED ISSUE #139: fn start should use `FN_LOG_LEVEL` rather than `GIN_MODE`
Mentioned most of this in https://github.com/fnproject/fn/issues/655 before closing it, but basically:

```
$ fn start --log-level DEBUG
can't create unix socket /var/run/docker.sock: device or resource busy
panic: gin mode unknown: DEBUG

goroutine 1 [running]:
github.com/fnproject/fn/vendor/github.com/gin-gonic/gin.SetMode(0xc420018029, 0x5)
	/go/src/github.com/fnproject/fn/vendor/github.com/gin-gonic/gin/mode.go:58 +0x156
..
```

```
"Env": [
                "GIN_MODE=DEBUG",
```

Help text is incorrect:

```
$ fn start --help
..
   --log-level value  --log-level DEBUG to enable debugging
..
```

And env var should be `FN_LOG_LEVEL`.

AstromechZA:
Fixes #139 and https://github.com/fnproject/fn/issues/655.

- help text was incorrect
- envvar passed to Docker container was incorrect

### DIFF
--- a/start.go
+++ b/start.go
@@ -19,14 +19,14 @@ func startCmd() cli.Command {
 		Flags: []cli.Flag{
 			cli.StringFlag{
 				Name:  "log-level",
-				Usage: "--log-level DEBUG to enable debugging",
+				Usage: "--log-level=debug to enable debugging",
 			},
 			cli.BoolFlag{
 				Name:  "detach, d",
 				Usage: "Run container in background.",
 			},
 			cli.StringFlag{
-				Name: "env-file",
+				Name:  "env-file",
 				Usage: "Path to Fn server configuration file.",
 			},
 		},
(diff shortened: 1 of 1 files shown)


start.go
```diff
@@ -19,14 +19,14 @@ func startCmd() cli.Command {
 		Flags: []cli.Flag{
 			cli.StringFlag{
 				Name:  "log-level",
-				Usage: "--log-level DEBUG to enable debugging",
+				Usage: "--log-level=debug to enable debugging",
```
treeder: Can we drop the equal sign here?

start.go
```diff
@@ -46,9 +46,9 @@ func start(c *cli.Context) error {
 		"-p", "8080:8080",
 	}
 	if c.String("log-level") != "" {
-		args = append(args, "-e", fmt.Sprintf("GIN_MODE=%v", c.String("log-level")))
+		args = append(args, "-e", fmt.Sprintf("FN_LOG_LEVEL=%v", c.String("log-level")))
 	}
-	if c.String("env-file") != ""{
```
treeder: OT: @denismakogon looks like you're not running `go fmt`.

start.go
AstromechZA: :+1: done

start.go
denismakogon: we should put that into CI, but yeah, have fmt checks for fn repo in pre-commit hooks.

treeder: Regardless of CI, it looks like your fmt isn't working.

PR #142, opened 2018-01-09
WIP: Can specify fn branch to build from with --fn-branch

treeder:
Requires https://github.com/fnproject/fn/pull/664

### DIFF
--- a/build_server.go
+++ b/build_server.go
@@ -205,7 +205,7 @@ RUN cd $D && dep ensure
 RUN cd $D && go build -o fnserver && cp fnserver /tmp/
 
 # final stage
-FROM fnproject/dind
+FROM fnproject/dind:17.12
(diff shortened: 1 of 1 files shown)


treeder:
Holding off on merging, want to measure time on this, `dep ensure` seems really slow...

treeder:
This doesn't work right. For some reason using `go get` (like current master) works, but using `dep ensure` does not, this happens:

```
INFO[0000] Setting log level to                          level=info
INFO[0000] started tracer                                url=
INFO[0000] datastore dialed                              datastore=sqlite3 max_idle_connections=256
FATA[0000] couldn't connect to docker daemon             error="Get http://unix.sock/_ping: dial tcp: lookup unix.sock on 192.168.65.1:53: no such host"
```

rikgibson:
No activity since Jan 10. Closing...

PR #147, opened 2018-01-22
cli: fn start does not use dind, but does not disable entrypoint

skinowski:
The stock entry point script attempts to start dockerd, which
fails with:

can't create unix socket /var/run/docker.sock: device or resource busy.

Instead, we override entrypoint to avoid dockerd start (disable dind).

### DIFF
--- a/start.go
+++ b/start.go
@@ -44,6 +44,7 @@ func start(c *cli.Context) error {
 		"-v", "/var/run/docker.sock:/var/run/docker.sock",
 		"--privileged",
 		"-p", "8080:8080",
+		"--entry-point ./fnserver",
 	}
 	if c.String("log-level") != "" {
 		args = append(args, "-e", fmt.Sprintf("FN_LOG_LEVEL=%v", c.String("log-level")))


rdallman:
ah, that makes sense. LGTM. I'm not sure of the test coverage in the CLI, did you happen to smoke test a task with `fn start` ?

skinowski:
yup, tested if with fn start (and verified env-file env vars are being passed to fn), my only concern here is hardcoding ./fnserver here, which is fragile a bit...

rdallman:
yea, if we have a test for `fn start` (eesh) then it would blow up pretty spectacularly if we change the dind image

treeder:
Does this mean if we run the image without the CLI, we need to pass in the `--entrypoint` flag?  Because if so, we should make it so we don't have to.

skinowski:
No, this only overrides the existing entrypoint, which is what -v /var/run/docker.sock:/var/run/docker.sock is trying to do.

skinowski:
Fixed by https://github.com/fnproject/cli/pull/147

PR #148, opened 2018-01-22
fn: fix typo docker-run args

skinowski:


start.go
```diff
@@ -44,7 +44,7 @@ func start(c *cli.Context) error {
 		"-v", "/var/run/docker.sock:/var/run/docker.sock",
 		"--privileged",
 		"-p", "8080:8080",
-		"--entry-point ./fnserver",
```
skinowski: I seriously need some coffee.. ^

denismakogon: Works for me
```
docker run --help | grep entry
      --entrypoint string              Overwrite the default ENTRYPOINT of the image
```
```
docker version
Client:
 Version:	17.12.0-ce
 API version:	1.35
 Go version:	go1.9.2
 Git commit:	c97c6d6
 Built:	Wed Dec 27 20:03:51 2017
 OS/Arch:	darwin/amd64

Server:
 Engine:
  Version:	17.12.0-ce
  API version:	1.35 (minimum version 1.12)
  Go version:	go1.9.2
  Git commit:	c97c6d6
  Built:	Wed Dec 27 20:12:29 2017
  OS/Arch:	linux/amd64
  Experimental:	true
```

PR #152, opened 2018-01-30
Add the config values from app.yaml on --deploy-all if there is not value for the config for that func.

hibooboo2:


deploy.go
```diff
@@ -182,6 +182,16 @@ func (p *deploycmd) deployAll(c *cli.Context, appName string, appf *appfile) err
 				ff.Path = p2
 			}
 		}
+		if appf.Config != nil {
```
denismakogon: Isn't this a piece of repeated code?

hibooboo2: What do you mean from where?

hibooboo2: The next line is checking func file config this is checking app file config. If we have no app file config which is appf then we do not need to add config to the func file. And the next one is checking if there is a func file config and if there is not initializing it to a map since we cannot assign values into a nil map.

denismakogon:
dummy CI, i retriggered build, hopefully, it's gonna be green soon

PR #154, opened 2018-02-01
Dont treat registry ports badly when cleaning image name during Bump

### LINKED ISSUE #151: cli doesn't do so hot with _real_ registries
not fake registries, which are just a docker hub username, this distinction shouldn't exist but it does atm. moving on...

this issue will contain known issues around this, which are numerous, will update this ticket. 

*** func.yaml name with a registry url in it is not properly bumped and breaks func.yaml (related to registry being a username being ambiguous, slightly, since subsequent deploys yell about it only)

```
 ✗: cat func.yaml                                                                                                                                                                                                                
name: localhost:5000/rdallman/hotyodawg                                                                                                                                                                                                       
version: 0.0.27                                                                                                                                                                                                                               
runtime: docker                                                                                                                                                                                                                               
format: http                                                                                                                                                                                                                                  
hello(master) ✗: cli deploy --app myapp
```

->

```
✗: cat func.yaml
name: localhost
version: 0.0.28
runtime: docker
format: http
```

AstromechZA:
Fixes #151;

Fixed the cleanImageName function here.

### Behaviour before:

```
$ cat func.yaml
name: localhost:5000/rdallman/hotyodawg
version: 0.0.3
runtime: go
entrypoint: ./func
$ ../fn bump
Bumped to version 0.0.4
$ cat func.yaml
name: localhost
version: 0.0.4
runtime: go
entrypoint: ./func
```

### Behaviour after:

```
$ cat func.yaml
name: localhost
version: 0.0.4
runtime: go
entrypoint: ./func
$ ../fn bump
Bumped to version 0.0.5
$ cat func.yaml
name: localhost:5000/rdallman/hotyodawg
version: 0.0.5
runtime: go
entrypoint: ./func
```

### DIFF
--- a/bump.go
+++ b/bump.go
@@ -135,10 +135,14 @@ func bumpVersion(funcfile *funcfile, t VType) (*funcfile, error) {
 	return funcfile, nil
 }
 
+// cleanImageName is intended to remove any trailing tag from the image name
+// since the version field conveys this information. More cleanup could be done
+// here in future if necessary.
 func cleanImageName(name string) string {
-	if i := strings.Index(name, ":"); i != -1 {
-		name = name[:i]
+	slashParts := strings.Split(name, "/")
+	l := len(slashParts) - 1
+	if i := strings.Index(slashParts[l], ":"); i > -1 {
+		slashParts[l] = slashParts[l][:i]
 	}
-
-	return name
+	return strings.Join(slashParts, "/")
 }
--- a/bump_test.go
+++ b/bump_test.go
@@ -60,3 +60,22 @@ func verifyVersion(tmp, version string) error {
 	}
 	return nil
 }
+
+func TestCleanImageName(t *testing.T) {
+	testCases := []struct {
+		input    string
+		expected string
+	}{
+		{"someimage:latest", "someimage"},
+		{"repository/image/name:latest", "repository/image/name"},
+		{"repository:port/image/name:latest", "repository:port/image/name"},
+	}
+	for _, c := range testCases {
+		t.Run(c.input, func(t *testing.T) {
+			output := cleanImageName(c.input)
+			if output != c.expected {
+				t.Fatalf("Expected '%s' but got '%s'", c.expected, output)
+			}
+		})
+	}
+}


rdallman:
thanks for taking this one. sorry not to mention this earlier, fwiw we use https://github.com/fnproject/fn/blob/master/api/agent/drivers/driver.go#L277 in the server side to do this parsing. it would be nice if we used the same code in the CLI, there seem to be numerous places where parsing images is buggy in the CLI but works fine in the server. as far as battle plan there for using that code since we aren't mono repo anymore, I'm not sure, I would like to avoid importing `fn` wholesale into the cli (it has come up for other things). at least, there's no good plan there for a common import, but it seems wise to copy the code over for image parsing in the CLI since we've diverged and the server version seems to work well. thoughts?

AstromechZA:
@rdallman I agree that it would be wise to pick up the same image parsing functions in the cli. I think copying over those functions with some appropriate table-tests would be useful here. By making it obvious that those functions are present in the cli we may drive more adoption of its usage. Incoming update.

AstromechZA:
Hilariously you can find a bunch of image name parsing by searching for `"/"`

AstromechZA:
@rdallman hmm, introducing those functions as specified in fn server will actually change functionality here in some cases: 

- an image name of `something` would be converted into `library/something`
- this may have unexpected results when combined with FN_REGISTRY
- I find it hard to tell whether the `name` field in the func.yaml is meant to be the function name or the destination image to push to. it's not really clear.

I think we should bring those functions on board when your proposal of FN_REGISTRY_URL is ready. Since that standardises the image name building greatly.

rdallman:
darn, thanks for looking. that makes sense, we have an added ambiguity for `library/` images conflicting with registries that should not have FN_REGISTRY prepended (further building the case to modify semantics). 

> I find it hard to tell whether the name field in the func.yaml is meant to be the function name or the destination image to push to. it's not really clear.

yea, I agree it's confusing. there's an issue filed https://github.com/fnproject/fn/issues/310 but since closed with some context that may be of interest, if you have any opinion would be glad to hear it.

this code looks good to fix this bug, let's ship it and fix image/registry stuff altogether elsewhere. thanks!

Issue #103, opened 2017-11-30
Functions with upper case in their names return an error


tteggel:
On Windows

mkdir helloNode && cd helloNode
fn init --runtime node
fn deploy ...
fn call ...

Actual: 
invalid argument "yanongena/helloNode:0.0.1" for t: invalid reference format: repository name must be lowercase

Expeced:
Hello World!

AstromechZA:
Quick reproduce attempt:

```
$ mkdir helloNode && cd helloNode
$ fn init --runtime python
Runtime: python
func.yaml created.
$ fn --verbose deploy --app hellonode
Deploying helloNode to app: hellonode at path: /helloNode
Bumped to version 0.0.2
Building image helloNode:0.0.2
invalid argument "helloNode:0.0.2" for t: invalid reference format: repository name must be lowercase
See 'docker build --help'.

Error during build. Run with `--verbose` flag to see what went wrong. eg: `fn --verbose CMD`
ERROR: error running docker build: exit status 125
$ fn --verbose deploy --app helloNode
Deploying helloNode to app: helloNode at path: /helloNode
Bumped to version 0.0.3
Building image helloNode:0.0.3
invalid argument "helloNode:0.0.3" for t: invalid reference format: repository name must be lowercase
See 'docker build --help'.

Error during build. Run with `--verbose` flag to see what went wrong. eg: `fn --verbose CMD`
ERROR: error running docker build: exit status 125
```

So the dockerPush attempt is throwing the error straight from the docker server itself. 
We'll need to add a bit of validation at `init` time and at `deploy` time.

PR #155, opened 2018-02-05
Upper case validation

reclaro:
This change checks that the name of the function passed as parameter with the `--name` option is lowercase.
If a name is not specified the default function name will be the (lowercase) name of the current folder.

### DIFF
--- a/init.go
+++ b/init.go
@@ -40,7 +40,7 @@ func initFlags(a *initFnCmd) []cli.Flag {
 	fgs := []cli.Flag{
 		cli.StringFlag{
 			Name:        "name",
-			Usage:       "name of the function. Defaults to directory name.",
+			Usage:       "name of the function. Defaults to directory name in lowercase.",
 			Destination: &a.ff.Name,
 		},
 		cli.BoolFlag{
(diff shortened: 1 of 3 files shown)


reclaro:
@hibooboo2 can you  help here with the CLA failure? Now I am part of the Oracle Organisation, how do I re-trigger the check again? fake commit?

init.go
```diff
@@ -197,15 +197,29 @@ func (a *initFnCmd) bindRoute(rt *models.Route) {
 	}
 }
 
+// validateFuncName checks if the func name is valid, the name can't contain a colon and
+// must be all lowercase
+func validateFuncName(name string) error {
```
denismakogon: this doesn't solve a problem completely, take a look at this https://github.com/fnproject/fn/blob/master/api/models/app.go#L30, the colon is not the only invalid char for the app name

reclaro: Hi Denis thanks for the review.
I am not sure I get your concern, I am trying to fix this issue: https://github.com/fnproject/cli/issues/103 which regards the name of the docker image created, the code you pointed me out refers to the validation of App name, am I right?
The validation rules for the App name are not a strict constraints for image name, for example I can have an image name longer than 30 chars.

denismakogon: Indeed, uppercase names is not the only one possible case when an app name is considered as invalid. Okay,
I see, had wrong context. LGTM.

PR #164, opened 2018-02-14
Lower sleep and use tmp file first

### LINKED ISSUE #37: Aborting fn install process leaves partially copied fn cli tool in target directory
```
curl -LSs https://raw.githubusercontent.com/fnproject/cli/master/install | sh
If you are just upgrading your functions cli client, ignore this and wait a few seconds.
You may press Ctrl+C now to abort this process.

(wait for download to start) 
(hit Ctrl+C)

fn 
[1]    4299 bus error (core dumped)  fn
```

installer should download to a temp location and move over fn when complete

hibooboo2:
closes #37

### DIFF
--- a/install
+++ b/install
@@ -52,18 +52,20 @@ url='https://github.com/fnproject/cli/releases/download'
 # perform some very rudimentary platform detection
 case "$(uname)" in
   Linux)
-    $sh_c "$curl /usr/local/bin/fn $url/$version/fn_linux"
+    $sh_c "$curl /tmp/fn_linux $url/$version/fn_linux"
+    $sh_c "$mv /tmp/fn_linux /usr/local/bin/fn"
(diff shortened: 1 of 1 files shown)


install
```diff
@@ -52,18 +52,20 @@ url='https://github.com/fnproject/cli/releases/download'
 # perform some very rudimentary platform detection
 case "$(uname)" in
   Linux)
-    $sh_c "$curl /usr/local/bin/fn $url/$version/fn_linux"
+    $sh_c "$curl /tmp/fn_linux $url/$version/fn_linux"
+    $sh_c "$mv /tmp/fn_linux /usr/local/bin/fn"
```
rbradford: Erroneous $ in front of mv here which breaks this script on Linux.

PR #165, opened 2018-02-14
Start fn in the background by default add --fg

### LINKED ISSUE #162: Make fn start run in background by default
Fn start starts fn in the foreground - 

On the whole users shouldn't care about the fn log output unless they are diagnosing an Fn problem - so I think the sensible default is to start in the background and support a --fg (vs -d) command 

This also makes the simplest user demo flow a little bit slicker as it keeps it in the same terminal

hibooboo2:
closes #162

### DIFF
--- a/start.go
+++ b/start.go
@@ -22,8 +22,8 @@ func startCmd() cli.Command {
 				Usage: "--log-level debug to enable debugging",
 			},
 			cli.BoolFlag{
-				Name:  "detach, d",
-				Usage: "Run container in background.",
+				Name:  "foreground, fg",
+				Usage: "Run container in the foreground.",
 			},
 			cli.StringFlag{
 				Name:  "env-file",
@@ -53,7 +53,7 @@ func start(c *cli.Context) error {
 	if c.String("env-file") != "" {
 		args = append(args, "--env-file", c.String("env-file"))
 	}
-	if c.Bool("detach") {
+	if !c.Bool("foreground") {
 		args = append(args, "-d")
 	}
 	args = append(args, functionsDockerImage)
--- a/test.sh
+++ b/test.sh
@@ -34,7 +34,7 @@ mkdir data
 touch data/fn.db data/fn.mq
 
 # start fn
-CONTAINER_ID=$($fn start -d)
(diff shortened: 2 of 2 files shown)


hibooboo2:
Easy patch all comes down to if we want this or not based on what @zootalures has said in the issue.

start.go
```diff
@@ -22,8 +22,8 @@ func startCmd() cli.Command {
 				Usage: "--log-level debug to enable debugging",
 			},
 			cli.BoolFlag{
-				Name:  "detach, d",
-				Usage: "Run container in background.",
+				Name:  "foreground, fg",
```
denismakogon: I really prefer that stay as it is, it looks like the docker ones, so maybe adjust the logic to make fn run in the background, which is totally fine

treeder:
I agree with Denis, I don't think it should run in the background by default.

hibooboo2:
@zootalures

zootalures:
@carimura  @shaunsmith  - this was based on observations of demos of which I guess you do the most (doing a terminal dance, hitting ctrl+z getting fn internal logs in the terminal)

zootalures:
Could the Nay sayers provide some justification as to why? (as per the justification for pro  in the original issue?)

treeder:
> Could the Nay sayers provide some justification as to why? (as per the justification for pro in the original issue?)

It's just not the norm, I'd prefer to stick to what people expect. And the norm is better in my opinion so you know what's going on. If it just disappears in the background and you have no idea why it didn't startup or whatever, that's not a very good experience. Adding the -d flag to make it go into the background for demos or whatever, doesn't seem like that big of a hurdle.

carimura:
FWIW I actually prefer keeping it open in demos. Didn't even realize there was a -d option, but that'll work fine if I want that experience.

skinowski:
We've seen too many cases of failures in 'fn start' due to developer env. With default foreground, it's easier to see these issues right away.

rdallman:
the tribe has spoken? thanks for the implementation @hibooboo2 - we should have discussed more on the original issue, sorry

PR #167, opened 2018-02-14
Generate pythons' FDK boilerplate

### LINKED ISSUE #111: Generate FDK boilerplate for python
The CLI should generate boilerplate with a basic example on fn init (as per java) for go and python. 

* Should honor "FDK_VERSION" env option (or CLI option) to select FDK version explicitly 
* Should get latest version remotely and fix this in the func.yaml/code if not set
* Should include a basic "hello world" example
* Should include a fn test

reclaro:
CLI now generates a basic example on fn init.
The example includes a fn test.
Closes #111

### DIFF
--- a/langs/python.go
+++ b/langs/python.go
@@ -2,6 +2,9 @@ package langs
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


langs/python.go
```diff
@@ -64,6 +67,85 @@ func (h *PythonLangHelper) IsMultiStage() bool {
...
+	if err := ioutil.WriteFile(testFile, []byte(pythonTestBoilerPlate), os.FileMode(0644)); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+const (
+	helloPythonSrcBoilerplate = `
+import sys
```
denismakogon: This boilerplate is wrong for anything except default format. This type of the boilerplate works only for one particular case - non-fdk default format. The boilerplate should be generated with respect to https://github.com/fnproject/fdk-python. But that is not all, fdk-python would never support python2, so if the python runtime version is 2.x CLI should fail.

treeder:
It's OK for FDK's to just support a single hot format (preferably json). They don't need to support the default format at all (since FDK's don't provide much value on top of default format).

treeder:
I'm OK with dropping 2.0 support?  Any really good reason to keep it?

denismakogon:
Nope, there’s no point to keep that.

carimura:
also want to link to this one -- separate but related while in there: https://github.com/fnproject/cli/issues/194

denismakogon:
Sorry for overtaking this one. Closing per https://github.com/fnproject/cli/pull/195

PR #171, opened 2018-02-15
Initial support for node boilerplate, better local dev experience

zootalures:
added basic boilerplate - this just resolves to a major version now  - ("0.x") (not great) 

supports locally vendored node_modules, and also copies package-lock.json over if it's there.

### DIFF
--- a/langs/node.go
+++ b/langs/node.go
@@ -1,5 +1,11 @@
 package langs
 
+import (
+	"os"
+	"path/filepath"
+	"io/ioutil"
+)
+
 type NodeLangHelper struct {
 	BaseHelper
 }
@@ -16,7 +22,7 @@ func (lh *NodeLangHelper) LangStrings() []string {
 }
 func (lh *NodeLangHelper) Extensions() []string {
 	// this won't be chosen by default
-	return []string{}
+	return []string{".js"}
 }
 
 func (lh *NodeLangHelper) BuildFromImage() (string, error) {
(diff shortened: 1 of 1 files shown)


langs/node.go
```diff
@@ -26,13 +32,59 @@ func (lh *NodeLangHelper) RunFromImage() (string, error) {
...
+
+const packageJsonContent = "{\n" +
+	"   \"name\": \"hellofn\", \n" +
+	"   \"version\": \"1.0.0\", \n" +
+	"   \"description\": \"example function\",\n" +
+	"   \"main\": \"func.js\",\n" +
+	"   \"author\": \"\",\n" +
+	"   \"license\": \"Apache-2.0\", \n" +
+	"   \"dependencies\": { \n" +
+	"      \"@fnproject/fdk\": \"0.x\"\n " +
```
denismakogon: Maybe put here a valid version?

langs/node.go
```diff
@@ -26,13 +32,59 @@ func (lh *NodeLangHelper) RunFromImage() (string, error) {
...
+	if err != nil {
+		return err
+	}
+	return ioutil.WriteFile(pathToFuncJs, []byte(funcJsContent), os.FileMode(0644))
+}
+
+func (lh *NodeLangHelper) HasBoilerplate() bool { return true }
+
+func (lh *NodeLangHelper) DefaultFormat() string { return "json" }
+
```
denismakogon: Does node FDK supports json?

langs/node.go
zootalures: That is a valid qualifier  and if you run npm install out side of the function it will lock to the version it got. 

We found with java that having to bump the CLI for every version caused more problems then having the version vary of its own free will. 

In an ideal world i would have had time to query npm for the latest version to fix (which gives you the best of both worlds) but didn't have time.

langs/node.go
zootalures: yes

langs/node.go
denismakogon: Okay, just tried that, you are right, it works.

langs/node.go
denismakogon: Cool

PR #173, opened 2018-02-20
Misc kotlin cleanup

delabassee:
- renamed kotline.go (!?!) to kotlin.go
- deleted a useless function call in the boilerplate function


treeder:
Thanks!

Issue #141, opened 2018-01-08
Use go dep tool during build


treeder:
Same way we're doing with other languages.

PR #178, opened 2018-02-23
Go: use vendor or dep tool during builds

denismakogon:
Closes: #141

### DIFF
--- a/langs/go.go
+++ b/langs/go.go
@@ -35,17 +35,20 @@ func (lh *GoLangHelper) RunFromImage() (string, error) {
 func (h *GoLangHelper) DockerfileBuildCmds() []string {
 	r := []string{}
 	// more info on Go multi-stage builds: https://medium.com/travis-on-docker/multi-stage-docker-builds-for-creating-tiny-go-images-e0e1867efe5a
-	// For now we assume that dependencies are vendored already, but we could vendor them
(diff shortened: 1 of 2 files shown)


langs/go.go
```diff
@@ -35,17 +35,20 @@ func (lh *GoLangHelper) RunFromImage() (string, error) {
...
+	if !vendor {
+		if exists("Gopkg.toml"){
+			r = append(r, "RUN go get -u github.com/golang/dep/cmd/dep",
+				"RUN cd /go/src/func/ && dep ensure -v")
+		}
+	}
+
 	r = append(r, "RUN cd /go/src/func/ && go build -o func")
+	// get rid of vendor after binary executable creation
+	r = append(r, "RUN cd /go/src/func/ && rm -fr vendor/")
```
treeder: This isn't necessary in multi-stage.

denismakogon: done

denismakogon:
Okay, i merged PR for fdk-go to get the example app from there, starting this moment CI should be green (at least locally it's happy with go func tests).

I feel a bit bad about not including glide to this PR, but it requires to include glide tool to the `fnproject/go:dev`. Will open PR for that.

treeder:
Don't worry about glide, dep seems to be taking it's place.

denismakogon:
Wish kubernets knew that =/

PR #181, opened 2018-02-27
Improved Java lang helper

delabassee:
`fn init --runtime java duke` now
- updates the pom.xml to include the func name as the artifactId
- generates boilerplate based on the func name (DukeFunction & DukeFunctionTest)
Note that capitals will be lowercased except the initial one (DukeDuke -> DukedukeFunction)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,74 @@
 
+# Contributing to Fn CLI
 
-Run `make install` to build and install to local machine for easy testing. 
+We welcome all contributions
+
+## How to contribute
+
+1. Fork the repo
+2. Fix an issue or create an issue and fix it
+3. Create a Pull Request that fixes the issue
+4. Sign the [CLA](http://www.oracle.com/technetwork/community/oca-486395.html)
+5. Once processed, our CLA bot will automatically clear the CLA check on the PR
+6. Good Job! Thanks for being awesome!
+
+## Advice for Forkers ##
+
+We do not advise cloning from your own fork (see [Getting the Repository](#getting-the-repository) below).
+
+Instead we suggest:
+1. Clone from the original repo
+2. Add your fork as a remote
+3. Push your changes to your fork
+4. Create your Pull Requests as normal
+
+## Documentation
+
+When creating a Pull Request, make sure that you also update the documentation
+accordingly.
+
+Most of the time, when making some behavior more explicit or adding a feature,
+documentation update is necessary.
+
+You will either update a file inside docs/ or create one. Prefer the former over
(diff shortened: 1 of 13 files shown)


zootalures:
I like the approach here but does it make more sense to do a little bit more abstraction around "init options" rather than having them as a blind arg-passed string slice? 

I'm worried about polluting the positional arguments space and also about future/other   things and how that works (i.e. what if there are 5 options but you only want to set 1) 


So - my suggestion here is (at the top level) : 

* add  slice-based option to `fn init` (not sure on naming but `-D` or `--config` or similar) 
* gather all options into a struct/map in the top-level CLI
* pass them down as config/context to the lang helper (this can be on construct I think) 

e.g.: 
```
fn init --runtime java --config pom.name=foo --config pom.groupId=mygroup  myfn 
```

then also perhaps  (maybe later) add a method to the lang helper that provides some docs for those args: 
```
fn init --runtime java --help 

...
usage : fn init --runtime java --config <key>=<val>
... 
Java config options  
pom.name : artifact name
pom.group: arifact group
....
```

delabassee:
I was trying to be as less intrusive as possible with this change as I just wanted to solve the (i.e. my) generic POM issue. But adding options would then allow externalizing (eg. Github) the boilerplate template as it was recently suggested.

Do you think that this is worth to do? I just want to avoid spending time on something that I'll be the only one to use.

zootalures:
Sorry I should have been a bit clearer on expectations here: 

* Supporting overriding the class names seems sensible, but 
* Changing the default name of the entails a bunch of changes to docs and tutorials 

So can we preserve the default behaviour (unless you can also go and  update all the docs/tutorials in fdk-java and tutorials) and add this as an option at least.

delabassee:
So I take care of updating docs & tutorials, you would be OK to have this as the default behavior? 
And we drop the " --config <key>=<val>" idea for now (btw, --config is already used for routes I think) ?

zootalures:
I'd prefer it if we had a flag of some kind I think and preserved the current behaviour by default.

zootalures:
Can you pick this up with @shaunsmith ? He's got a handle on the impact to docs/tutorials (sorry for confusion)

shaunsmith:
Let's roll these requirements into an effort that is kicking off to bring consistent treatment and behaviour for all of the core supported languages.  We already have a problem with diverse levels of support for each language so rather than increase the diversity let's define general requirements for boilerplate generation and the FDK's that are being leveraged.

rikgibson:
With the externalising of boilerplate generation into customisable init-images, seems this PR is redundant now - any objection to it being closed?

delabassee:
With the _init-image_ approach this PR is not needed anymore!

Issue #182, opened 2018-02-27
missing required header: fn_method


shaunsmith:
CLI 0.4.53 is throwing an error when calling `fn run` on boilerplate generated Java application.  

```
Building image hello:0.0.2 .
Incoming HTTP frame is missing required header: fn_method

ERROR: exit status 2
```

func.yaml
```
name: hello
version: 0.0.2
runtime: java
cmd: com.example.fn.HelloFunction::handleRequest
build_image: fnproject/fn-java-fdk-build:jdk9-1.0.56
run_image: fnproject/fn-java-fdk:jdk9-1.0.56
format: http
```

@owencliffe commented:
```
looks like a regression/bug  in the new HTTP support in the cli - bug it (edited)
(previously fn run only called with default contract - now does HTTP, looks to be inconsistent with what the server is sending
```

treeder:
It actually uses the format in the funcfile now, which it wasn't before for some reason. Will fix.

treeder:
@shaunsmith can you verify and close that it's fixed now. Thanks!

PR #183, opened 2018-02-27
Remove newline from HTTP (content-length delimits requests), set same headers as fn

zootalures:


run.go
```diff
@@ -220,21 +221,25 @@ func runff(ff *funcfile, stdin io.Reader, stdout, stderr io.Writer, method strin
 				return fmt.Errorf("error creating http request: %v", err)
 			}
 			req.Header.Set("Content-Type", contentType)
-			req.Header.Set("Fn_deadline", deadlineS)
+
+			req.Header.Set("FN_REQUEST_URL", LocalTestURL)
```
treeder: Should these by in the same case as here: https://github.com/fnproject/fn/pull/660#issuecomment-356157279 ?

zootalures:
fixes #182

run.go
zootalures: Pretty sure that header.Set canonicalises so don't think it matters here - just trying to be consistent with the headers in http.go

PR #189, opened 2018-03-03
Expanded build instructions for CLI and link from README

crush-157:
CONTRIBUTING.MD:
- Copied "how to contribute" from fn (core) CONTRIBUTING.md
- Added detailed instructions on how to build from source

README.md
- Added link to CONTRIBUTING.md for build from source instructions.

PR #190, opened 2018-03-05
updates to contrib

rdallman:
tiny follow ups to #189

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,21 +32,10 @@ asking for suggestions on how to address the documentation part.
 
 ### Cloning the Repository ###
 
-Whether cloning from your Fn CLI repo or your own fork, you need to take care with the location.
+`$ go get github.com/fnproject/cli`
 
-`$ mkdir $GOPATH/src/github.com/fnproject/cli `
-
-`$ cd $GOPATH/src/github.com/fnproject/cli `
-
-Then you can either
-
-1. Clone from the Fn CLI repo:
-
-	`$ git clone https://github.com/fnproject/cli.git`
-
-2. Clone from your own fork:
-
-	`$ git clone git@github.com:<YOUR-GITHUB_USERNAME>/cli.git`
+Note that we use `dep` for dependencies, and this will not do an installation.
+See below.
 
 ### Building ###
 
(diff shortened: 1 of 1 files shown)


treeder:
I think developers should go the `git clone` route so they know what's going on. The reason for this in the first place was to help people that aren't familiar with Go and this change doesn't help with that. eg: Where the hell did `go get` put this stuff?

rdallman:
i'll acquiesce - I will reduce it to checking out of `github.com/fnproject/fn` instead of giving the option for a fork build. multiple examples from other people with more stars than us: https://github.com/kubernetes/kubernetes https://github.com/gin-gonic/gin https://github.com/gohugoio/hugo https://github.com/coreos/etcd/blob/master/Documentation/dl_build.md#build-the-latest-version

rdallman:
updated https://github.com/fnproject/cli/pull/190/commits/9e6822ab44cde1af69d3767c9fcc99a54beac52d @treeder

crush-157:
@rdallman the reason I included the checking out from a fork is because that's what someone external to the core team who wants to contribute (like me) is going to do most of the time.

rdallman:
> @rdallman the reason I included the checking out from a fork is because that's what someone external to the core team who wants to contribute (like me) is going to do most of the time.

Yep, it is a problem. I guess my hope is that we can steer people towards:

* clone `fnproject`'s repo
* add forked repo as a remote

I reviewed a number of go projects https://github.com/fnproject/cli/pull/190#issuecomment-370591994 and none of them explicitly mention pushing to a fork branch to submit a PR in their 'get the code and build it' step -- I do agree it seems nice not to muddle these two. If we recommend to get the code by cloning to the proper location it will avoid build errors (as experienced), at least, and then it's easy to go from there to contribute rather than having to move things around :)

there are, of course, likely a number of ways to accomplish this.

treeder:
Cloning then adding remote makes the most sense. Which maybe makes the `go get` way easier... :) 

Also, I didn't notice the follow up that already had the person cd'ing into the right dir, was just looking at the changes instead of the full file (my bad), so if you want to go back to `go get`, go for it.

rdallman:
> Also, I didn't notice the follow up that already had the person cd'ing into the right dir, was just looking at the changes instead of the full file (my bad), so if you want to go back to go get, go for it.

there seems to be a mix among the community. since the cli is not build-able without `make dep`, clone might make more sense, since `go get` makes some installation assumptions.

PR #196, opened 2018-03-07
Advice for forkers

crush-157:
Just added some advice to forkers explaining about cloning from the fnproject/cli repo and adding their fork as an additional remote, rather than cloning from their own and adding the original as an upstream remote.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,75 @@
 
+# Contributing to Fn CLI
 
-Run `make install` to build and install to local machine for easy testing. 
+We welcome all contributions
+
+## How to contribute
+
+1. Fork the repo
+2. Fix an issue or create an issue and fix it
+3. Create a Pull Request that fixes the issue
+4. Sign the [CLA](http://www.oracle.com/technetwork/community/oca-486395.html)
+5. Once processed, our CLA bot will automatically clear the CLA check on the PR
+6. Good Job! Thanks for being awesome!
+
+## Documentation
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -12,6 +12,16 @@ We welcome all contributions
...
 6. Good Job! Thanks for being awesome!
 
+## Advice for Forkers ##
+
+We do not advise cloning from your own fork (see [Getting the Repository](#Getting_the_Repository) below).
+
+Instead we suggest:
+1. Clone from the original repo
+2. Add your fork as a remote
+3. Push your changes your fork
```
treeder: typo: "to your fork"

CONTRIBUTING.md
```diff
@@ -30,7 +40,7 @@ asking for suggestions on how to address the documentation part.
 - [Go](https://golang.org/doc/install)
 - [Dep](https://github.com/golang/dep)
 
-### Getting the Repository ###
+### Getting the Repository <a name="Getting_the_Repository"></a> ###
```
treeder: GitHub already adds anchors to all the titles, so this is unnecessary.

CONTRIBUTING.md
crush-157: Thanks.

CONTRIBUTING.md
crush-157: Noted.  Thanks

crush-157:
Changes made

PR #207, opened 2018-03-14
added --no-bump option for deploy

MartinWeindel:
To support external version management, bumping of version can be disabled using flag --no-bump.


carimura:
@MartinWeindel looks like the check for CLA is passed, but the block is due to missing reviewer. Let's see if we can push this through today.

Issue #211, opened 2018-03-23
[Go] run image doesn't work well with very specific golang code


denismakogon:
Original error: https://github.com/fnproject/dockers/issues/12

denismakogon:
Will remove new `COPY` statement as soon as new `fnproject/go` will be released. PR for that: https://github.com/fnproject/dockers/pull/13

PR #212, opened 2018-03-24
Add timezone database copy for go run image

denismakogon:
- the size of the images will grow for 355Kb

Closes: #211


langs/go.go
```diff
@@ -65,6 +65,8 @@ func (h *GoLangHelper) DockerfileBuildCmds() []string {
 
 func (h *GoLangHelper) DockerfileCopyCmds() []string {
 	return []string{
+		"COPY --from=build-stage /usr/local/go/lib/time/zoneinfo.zip " +
```
treeder: This is only a Go issue?

denismakogon: So, i'm not sure, but faced with this while was developing a function with go

rdallman: we should just have it in base image we build from? string of issues over this, made some suggestions https://github.com/fnproject/dockers/pull/13 - we probably need to add for all langs (like certs)

PR #214, opened 2018-03-29
don't extract message from 'default' errors

rdallman:
the swagger error handling is not very fun, default errors are not necessarily
able to parse certain errors into each errors' 'default' type. in theory, we
should be able to `return err` but we would get errors that look like this:

```
return fmt.Sprintf("[PUT /apps/{app}/routes/{route}][%d] PutAppsAppRoutesRoute default  %+v", o._statusCode, o.Payload)
```

so the compromise seems to leave us to handle specific errors in the cli
itself, and let default errors print the uglier, more detailed blob like
above, since the error types themselves only implement `Error() string` and
`Code() int` and don't allow us to get at the `Payload` field, which is
sometimes nil, itself. looked for 15 minutes and couldn't find a way to do
this, at least. for the more interesting errors, we probably want to get the
code and the full body / route anyway, to help debug.

closes https://github.com/fnproject/fn/issues/907

### DIFF
--- a/apps.go
+++ b/apps.go
@@ -133,13 +133,8 @@ func (a *appsCmd) list(c *cli.Context) error {
 			switch e := err.(type) {
 			case *apiapps.GetAppsAppNotFound:
 				return fmt.Errorf("%v", e.Payload.Error.Message)
-			case *apiapps.GetAppsAppDefault:
-				return fmt.Errorf("%v", e.Payload.Error.Message)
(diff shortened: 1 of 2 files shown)


rdallman:
bump @denismakogon @treeder @skinowski easy one

rdallman:
bump 2: the bumpening @denismakogon @treeder easy one

Issue #217, opened 2018-04-06
As an Fn user I want to define multiple deployment contexts and be able to easily switch between them.


rikgibson:
### Summary
A user should be able to use a local dev environment, or **Context**, for development and then be able to easily switch to a different context, e.g. dev/test/prod for deployment, such that any CLI commands that call an Fn Server API will be routed to the server pointed to by currently selected context, and configured with any required AuthN/AuthZ characteristics. Users should be able to create multiple contexts representing different deployment environments, with each context specifying an appropriate **Provider** to use and a set of properties required/understood by that Provider.

### CLI enhancements
- Create one or more remote deployment contexts:
  * `fn create context dev`
  * `fn create context test --key provider --value default`
- Populate the new context with a set of context-specific properties, e.g. type of cloud provider, Fn Server URL, target container registry and one or more cloud-provider specific properties, e.g.
  * `fn update context dev --key fn-api-url --value test.acme.com`
- List available contexts, showing all available contexts along with which is currently selected for use.
  * `fn list contexts`
- Switch to use a different context:
  * `fn use context test`
- The current context can be overridden on any CLI command using `--context=some-context`
  * `fn --context=test call myApp /hello`

### Contexts
Contexts are persisted in individual files underneath an Fn-specific directory, for example:

`~`
`  .fn`
`    config.yml`
`    contexts`
`      local.yml`
`      dev.yml`

The `config.yml` will contain the name of the currently selected context:

`- current-context : local`

The CLI will load the `config.yml` on launch and parse to determine the currently selected context, which will then also be loaded.  If the `.fn` directory or `config.yml` are not present, the CLI will create them and populate with simple default information.

### local.yml
`- context`
`    name : local`
`    provider : default`
`    fn-api-url : localhost`
`    registry : `

### Configuration Using Environment Variables
The current CLI implementation supports a couple of env vars - FN_API_URL and FN_REGISTRY.  These should be replaced by equivalent properties defined in contexts, however there are still use cases where the use of these variables is useful or preferable (CI testing for example). If the CLI detects both an env var and an equivalent context property, the env var will override the context property, with the CLI outputting a warning to this effect:

`Warning: environment variable FN_API_URL is overriding configured context property fn-api-url from context "local"`

### Providers
The effect of all of this is that when the CLI wants to make an outbound Fn API call, the context configuration bundle (including any env var overrides) should be passed into the Fn Go SDK such that it can be handed off to a relevant *Provider* to encode the outbound HTTP request accordingly.

[...]

PR #224, opened 2018-04-11
Migrating env vars config to viper

JadeRedworth:
Moved the creation of aliases after api_url is set within the init function in main.go due to the creation of the apiClient before viper had loaded the registry env var

Removed the hasRegistry interface in common.go and replaced with a commandArgOverride function in main.go

This is the first part of github issue #217

### DIFF
--- a/test/funcfile-docker-rt-tests/fn_test.go
+++ b/test/funcfile-docker-rt-tests/fn_test.go
@@ -4,23 +4,20 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
-	"log"
 	"os"
 	"os/exec"
 	"path"
 	"testing"
+
+	"github.com/spf13/viper"
 )
 
 func init() {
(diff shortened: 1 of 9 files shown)


test/funcfile-docker-rt-tests/fn_test.go
```diff
@@ -4,23 +4,20 @@ import (
...
 	dockerUser := os.Getenv("DOCKER_USER")
-	var err error
 	if dockerUser == "" {
 		dockerUser = "funcster"
 	}
-	err = os.Setenv("FN_REGISTRY", dockerUser)
-	if err != nil {
-		log.Fatalf("couldn't set env var: %v", err)
-	}
+	viper.Set("registry", dockerUser)
```
hibooboo2: Maybe we should get rid of this default user and only set registry if docker_user is there?

hibooboo2: Nvm this is for testing. Reading too fast. My bad.

PR #230, opened 2018-04-17
Add context configuration directory hierarchy and context switching

JadeRedworth:
Related to GitHub issue #217

Created a function to checks if .fn directory exists before commandArgOverrides tries to read in the config. If .fn directory does not exist, EnsureCongiuration will create it with the config.yaml,
contexts folder with a default.yaml context. Multiple checks are made for each section of the .fn directory.

Added a config.go file to handle the const and envvars and the predefined context maps.

Move the creation of the fn API client before each of the commands are created to enable the configuration file to be read in and the values set by viper. The --context flag provides the ability to switch between context files:
```
fn --context default apps create <app> ...
```

This has been tested by creating different fn servers one localhost:8081 and localhost:8082 and
switching between --context [ default | default2 ]

config.yaml
```
current-context: [ default | default 2 ]
```

default.yaml
```
api_url: https://localhost:8081
provider: default
registry: ""
```

default2.yaml
```
api_url: https://localhost:8082
provider: default2
registry: ""
```

Added the start of the context command structure to enable management of contexts.
_note: the context commands have not yet been created. Users can only switch between context with the `--context` flag and must use it on each command they run._


main.go
```diff
@@ -125,7 +140,7 @@ func init() {
 	viper.AutomaticEnv() // read in environment variables that match
 
 	viper.SetEnvPrefix("fn")
-	viper.SetDefault("api_url", "http://localhost:8080")
+	viper.SetDefault(envFnAPIURL, "http://localhost:8080")
 
 	// create aliases after api_url set
 	aliases = map[string]cli.Command{
```
mantree: These can go back to where aliases is defined now that the building of the commands doesn't create the http client.

context.go
```diff
@@ -0,0 +1,128 @@
...
+				ArgsUsage: "<context>",
+				Action:    set,
+			},
+		},
+	}
+}
+
+// EnsureConfiguration ensures context configuration directory hierarchy is in place, if not
+// creates it and the default context configuration files
+func EnsureConfiguration() {
```
mantree: Maybe this should be in the config.go file.

JadeRedworth: Agreed, also makes sense to potentially not include this context.go file as it provides no value for this PR if EnsureConfiguration() is moved?

config.go
```diff
@@ -0,0 +1,93 @@
...
+
+	homedir "github.com/mitchellh/go-homedir"
+	yaml "gopkg.in/yaml.v2"
+)
+
+const (
+	rootConfigPathName     = ".fn"
+	contextsPathName       = "contexts"
+	configName             = "config"
+	contextConfigFileName  = "config.yaml"
```
treeder: Is it one file per context or all the contexts in this one file?  And if all in one, should default also be in that same file?

JadeRedworth: Its one file per context with `default.yaml` being created initially.

PR #231, opened 2018-04-17
Making config importable

mantree:
Having the config names defined in the main package meant they
couldn't be reused in other packages, namely client/api.

This creates the config package to contain the config key definitions
and makes them exportable.

Also moved the loadConfiguration func into config from main, and it
feels like a better place for it to live.

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -5,15 +5,11 @@ import (
 	"net/url"
 	"os"
 
+	"github.com/fnproject/cli/config"
 	fnclient "github.com/fnproject/fn_go/client"
 	openapi "github.com/go-openapi/runtime/client"
 	"github.com/go-openapi/strfmt"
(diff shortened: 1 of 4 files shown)


client/api.go
```diff
@@ -26,21 +22,24 @@ func Host() string {
 }
 
 func HostURL() (*url.URL, error) {
-	apiURL := viper.GetString(envFnAPIURL)
+	apiURL := viper.GetString(config.EnvFnToken)
```
JadeRedworth: That needs to envFnAPIURL

mantree: nice one, thanks

mantree:
Recreating to trigger a check rerun

mantree:
Recreating to trigger a check rerun

PR #236, opened 2018-04-19
Support for CloudEvents format.

treeder:
Requires: https://github.com/fnproject/fn/pull/948

### DIFF
--- a/run.go
+++ b/run.go
@@ -16,10 +16,11 @@ import (
 )
 
 const (
-	DefaultFormat = "default"
-	HttpFormat    = "http"
-	JSONFormat    = "json"
-	LocalTestURL  = "http://localhost:8080/myapp/hello"
+	DefaultFormat    = "default"
+	HttpFormat       = "http"
+	JSONFormat       = "json"
+	CloudEventFormat = "cloudevent"
+	LocalTestURL     = "http://localhost:8080/myapp/hello"
 )
 
 func run() cli.Command {
@@ -263,6 +264,8 @@ func handle(format string) handlerFunc {
 		return handleHTTP
 	case JSONFormat:
 		return handleJSON
+	case CloudEventFormat:
+		return handleCloudEvent
 	default:
 		return handleDefault
 	}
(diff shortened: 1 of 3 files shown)


run.go
```diff
@@ -334,6 +337,23 @@ func handleJSON(conf runConfig) (io.Reader, io.Writer, error) {
 	return stdin, stdout, nil
 }
 
+func handleCloudEvent(conf runConfig) (io.Reader, io.Writer, error) {
+	fmt.Println("HANDLE CLOUD EVENT")
```
rdallman: rm

run.go
```diff
@@ -334,6 +337,23 @@ func handleJSON(conf runConfig) (io.Reader, io.Writer, error) {
...
+	var b strings.Builder
+	for i := 0; i < conf.runs; i++ {
+		body, err := createCloudEventInput(conf.callID, conf.contentType, conf.deadline, conf.method, conf.url, conf.stdin)
+		if err != nil {
+			return nil, nil, fmt.Errorf("error creating input: %v", err)
+		}
+		fmt.Println("Sending in: ", body)
```
rdallman: `s/^.*fmt.Println(.*$//`

run_cloud_event.go
```diff
@@ -0,0 +1,107 @@
...
+				d, err := json.Marshal(jsout.Data)
+				if err != nil {
+					fmt.Printf("Error marshalling function response 'data' to json. %v\n", err)
+					return
+				}
+				stdout.Write(d)
+			} else if jsout.ContentType == "text/plain" {
+				stdout.Write([]byte(jsout.Data.(string)))
+			} else {
+				fmt.Printf("Error: Unknown content type: %v\n", jsout.ContentType)
```
rdallman: `fmt.Fprintln(os.Stderr, ...)` ?

run_cloud_event.go
```diff
@@ -0,0 +1,107 @@
...
+			jsout := &CloudEvent{}
+			err = dec.Decode(jsout)
+			if err != nil {
+				fmt.Println("error decoding", err)
+				return
+			}
+			if jsout.ContentType == "application/json" {
+				d, err := json.Marshal(jsout.Data)
+				if err != nil {
+					fmt.Printf("Error marshalling function response 'data' to json. %v\n", err)
```
rdallman: fmt.Fprintln(os.Stderr, ...) ?

run_cloud_event.go
```diff
@@ -0,0 +1,107 @@
...
+				RequestURL: requestURL,
+				Type:       "http",
+				Headers: map[string][]string{
+					"Content-Type": {contentType},
+				},
+			},
+			"deadline": deadline,
+		},
+	}
+	fmt.Println("INPUT:", string(input))
```
rdallman: rm ?

run_cloud_event.go
```diff
@@ -0,0 +1,107 @@
...
+	} else if contentType == "application/json" {
+		d := map[string]interface{}{}
+		err = json.Unmarshal(input, &d)
+		if err != nil {
+			return "", fmt.Errorf("Error unmarshalling json input: %v", err)
+		}
+		in.Data = d
+	} else {
+		in.Data = string(input)
+	}
```
rdallman: still not sure about this. from the spec when i read it this really seems like a string, where the content type defines to the user how they should decode it. the spec is not clear about this, but it does say arbitrary and not string. 

the downside being that if we do this differently than everyone else (whichever way that is) that implements the spec then users are going to run into errors parsing this field. 

i guess for every other format except json (`application/xml`, `application/vnd.hzn-3d-crossword`, `application/x-7z-compressed`, etc etc) these are going to have to be a string so that we can wedge them into json? that seems weird in itself, to me.

rdallman:
LGTM w/ println cleanup here too. i smell `fnproject/event` repo coming on and i'm still confused about all the content type shenanigans but that can probably be glossed over since we can fix the glitch later.

run_cloud_event.go
rdallman: https://github.com/cloudevents/spec/blob/master/http-transport-binding.md#22-data-attribute

treeder:
Updated.

PR #237, opened 2018-04-19
Context management commands

JadeRedworth:
Related to issue: [#217](https://github.com/fnproject/cli/issues/217) 

Added the context commands, create, delete, list, set and unset to manage context files.

Create command checks the file hasn't already been created and creates it either with 'default'
values or with values passed in through the `--provider` and `--apiurl` flags.

Delete command checks the `<context>` files exists and that it is not the current context. Only then can it be deleted. The default context cannot be deleted as it is protected.

Set checks there is a context file and sets the current context to the passed in `<context>` and
updates the config.yaml file.

Unset will update the config.yaml to `""` un-setting the current-context.

List shows all available context files, unmarshals them and appends them to a table, showing the current context with a `*`

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -1,6 +1,6 @@
 # Context command
 
-Context files are used to define deplployment enviroments. Multiple contexts representing different enviroments can be created with each context specifying an appropriate provider  and a set of properties required/understood by that provider with the ability to switch between them. Contexts are persisted in individual files under the `.fn` directory which, if not present, is created on launch.
+Context files are used to define deployment enviroments. Multiple contexts representing different enviroments can be created with each context specifying an appropriate provider  and a set of properties required/understood by that provider with the ability to switch between them. Contexts are persisted in individual files under the `.fn` directory which, if not present, is created on launch.
 
 ```
 ~ 
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -43,19 +43,19 @@ $ fn context create <context>
 $ fn context create <context> --api-url foo --provider bar --registry <dockerhub-username>
 ```
 
-### Setting a Context 
+### Using a Context
 
-To set a context use `s` or `set`.
+To use a context use `u` or `use`.
 
 ```
-$ fn context set <context>
+$ fn context use <context>
 ```
 
 ### Deleting a Context 
  
 To delete a context use `d` or `delete`.
 
-_You can not delete the currently set context or the default context as it is protected._
+_You can not delete the currently used context or the default context as it is protected._
 
 ```
 $ fn context delete <context>
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -1,0 +1,74 @@
+# Context command
+
+Context files are used to define deplployment enviroments. Multiple contexts representing different enviroments can be created with each context specifying an appropriate provider  and a set of properties required/understood by that provider with the ability to switch between them. Contexts are persisted in individual files under the `.fn` directory which, if not present, is created on launch.
+
+```
+~ 
+  .fn
+     config.yaml
+     contexts
+        default.yaml
+```
+
+The `config.yaml` contains the name of currently selected context, when first created the current context is not set:
+
+ ```
+ current-context: ""
+ ``` 
+
+Within the `default.yaml` default values are set: 
+```
+api_url: http://localhost:8080/v1
+provider: default
+registry: ""
+```
+
+* api_url - the fn-server endpoint.
+* provider - the a specific provider which identifies a set of properties required/understood by that provider.
+* registry - the Docker registry username to push images to 
+[registry.hub.docker.com/`registry`].
+
+### Listing Contexts
+The `fn context` command accepts either `l` or `list` to view a list of contexts.
+```
+$ fn context list
+```
+
+### Creating a Context 
+
+To create a context use `c` or `create`. The context file will be created with default values but option flags can be used override.
+
+```
+$ fn context create <context>
+$ fn context create <context> --api-url foo --provider bar --registry <dockerhub-username>
+```
+
+### Setting a Context 
+
+To set a context use `s` or `set`.
+
+```
+$ fn context set <context>
+```
+
+### Deleting a Context 
+ 
+To delete a context use `d` or `delete`.
+
+_You can not delete the currently set context or the default context as it is protected._
+
+```
+$ fn context delete <context>
+```
+
+### Unsetting Context 
+
+To unset the current context use `unset`:
+
+```
+$ fn context unset
+```
+
+# Enviroment Variables
+
+_The current supported env vars 'FN_API_URL' and 'FN_REGISTRY' will override the configured context properties_.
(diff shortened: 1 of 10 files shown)


zootalures:
think you are missing the Gopkg.lock file with the tablewriter changes.

context.go
```diff
@@ -0,0 +1,209 @@
...
+						Name:  "config",
+						Usage: "context configuration",
+					},
+					cli.StringFlag{
+						Name:  "provider",
+						Usage: "context provider",
+					},
+					cli.StringFlag{
+						Name:  "apiurl",
+						Usage: "context api url",
```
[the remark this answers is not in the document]
JadeRedworth: Yeah we allow for command arg overrides as well, if they override the set the API_URL env var it will be used instead of what is set in the context file

context.go
```diff
@@ -0,0 +1,209 @@
...
+				Aliases:   []string{"c"},
+				Usage:     "create a new context",
+				ArgsUsage: "<context>",
+				Action:    create,
+				Flags: []cli.Flag{
+					cli.StringSliceFlag{
+						Name:  "config",
+						Usage: "context configuration",
+					},
+					cli.StringFlag{
```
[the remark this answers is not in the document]
JadeRedworth: Yes for sure, the next step would be when the provider defined was say 'oracle' we would create a context file with specific oracle fields. This will lead on from the work @mantree has just done.

context.go
```diff
@@ -0,0 +1,209 @@
...
+						Name:  "provider",
+						Usage: "context provider",
+					},
+					cli.StringFlag{
+						Name:  "apiurl",
+						Usage: "context api url",
+					},
+				},
+			},
+			{
```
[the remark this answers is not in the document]
JadeRedworth: The was we have it set up that before any commands it checks for the `./fn/contexts/` and if there is a default context file. If not it will create one, so deleting it is actually pointless. Should be protected 👍thanks.

context.go
```diff
@@ -0,0 +1,209 @@
...
+		fmt.Printf("ERROR: can not delete the current context: %v \n", context)
+		os.Exit(1)
+	}
+
+	os.Remove(createFilePath(context))
+	fmt.Printf("Context %v deleted \n", context)
+}
+
+func set(c *cli.Context) {
+	context := c.Args().Get(0)
```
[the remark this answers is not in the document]
JadeRedworth: @mantree what you recon about this one? Within the main.go we check that the config.yaml current-context is a valid context file. If not you can't do anything without manually changing the file..

client/api.go
```diff
@@ -27,14 +28,21 @@ func HostURL() *url.URL {
 }
 
 func hostURL(urlStr string) *url.URL {
+	apiUrl := os.Getenv("FN_API_URL")
+	currentContext := viper.Get(config.CurrentContext)
+	if apiUrl != "" && currentContext != "" {
+		fmt.Println("Warning: environment variable FN_API_URL is overriding configured context property api_url from current-context")
```
zootalures: Here and elsewherre I think errors/warnings should go to stderr: 

```
fmt.Fprintf(os.Stderr,....
```

rikgibson: +1 on routing warnings to stderr -  @shaunsmith would like the ability to output results of 'list' operations as JSON and parse the output, so a warning in there would pollute the output.

More broadly though, as commented elsewhere in #249, I wonder whether we should stop allowing the mixing of contexts and env vars - use one or the other. Leave context unset to use env vars, or use a context in which case env vars are ignored.

JadeRedworth: Would that require an `fn context unset` command in the case they have a context set but want to revert back using envvars? Or do we just expect them to edit the config.yaml to remove current-context?

config/config.go
```diff
@@ -35,105 +37,134 @@ const (
...
+}
+
 // EnsureConfiguration ensures context configuration directory hierarchy is in place, if not
 // creates it and the default context configuration files
 func EnsureConfiguration() {
-	home, err := homedir.Dir()
+	home, err := GetHomeDir()
 	if err != nil {
-		panic(err)
+		fmt.Printf("%v", err)
```
kojustin: is this supposed to print to stderr?

JadeRedworth: Oh yes thank you i missed one!

context.go
```diff
@@ -0,0 +1,288 @@
...
+
+	}
+	path, err := createFilePath(context)
+	if err != nil {
+		return fmt.Errorf("%v", err)
+	}
+	_, err = os.Create(path)
+	if err != nil {
+		return fmt.Errorf("%v", err)
+	}
```
kojustin: It probably doesn't matter since this program will exit , but this will leak the File object returned by `os.Create`.

But `config.WriteYamlFile` below will perform the file truncation and create as well, so this call to `os.Create` may not be necessary?

config/config.go
JadeRedworth: @denismakogon does `log.Println`/`log.Fatal` go to stdout or stderr?

hibooboo2: Yes but fatal panics.

client/api.go
```diff
@@ -27,14 +28,21 @@ func HostURL() *url.URL {
 }
 
 func hostURL(urlStr string) *url.URL {
+	apiUrl := os.Getenv("FN_API_URL")
```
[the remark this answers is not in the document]
hibooboo2: Just remove this actually. AS it is already picked up by viper.

CONTEXT.md
```diff
@@ -0,0 +1,74 @@
+# Context command
+
+Context files are used to define deplployment enviroments. Multiple contexts representing different enviroments can be created with each context specifying an appropriate provider  and a set of properties required/understood by that provider with the ability to switch between them. Contexts are persisted in individual files under the `.fn` directory which, if not present, is created on launch.
```
treeder: spelling `deplployment`.

CONTEXT.md
```diff
@@ -0,0 +1,74 @@
...
+$ fn context create <context>
+$ fn context create <context> --api-url foo --provider bar --registry <dockerhub-username>
+```
+
+### Setting a Context 
+
+To set a context use `s` or `set`.
+
+```
+$ fn context set <context>
```
treeder: Would `use` be a better name than `set`? Set feels like it should be setting/changing a value, like when it's used for config vars.

treeder: And I see there's no way to set the values in the context, so the user is expected to edit the files to change the context variables I assume?  Could `set` be used to set the vars?  eg: `fn context set api-url foo` ?

JadeRedworth: I like the idea of both `set` and `use`, although we do have an `unset` command which links in with `set`. Does it make sense to have a `use` and `unset`?

An update command is coming where users can `fn context update <key> <value>` but for now they will edit it manually in their file. Also for the default context they can pass in the value `--api-url, --registry and --provider`.

JadeRedworth:
Have address @treeder changes and will go ahead with merge as 2+ others have tested and approved.

PR #241, opened 2018-04-21
Add unit tests for fn api url configuration parsing

mantree:
Instead of running the full test suite multiple times, test the
HostURL parsing function with a set of unit tests.

Defaults for scheme (http), scheme appropriate port, and backwards
compatible path (/v1)

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -23,19 +23,32 @@ func Host() string {
 }
 
 func HostURL() *url.URL {
-	apiURL := viper.GetString(config.EnvFnAPIURL)
+	return hostURL(viper.GetString(config.EnvFnAPIURL))
+}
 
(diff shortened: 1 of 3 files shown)


client/api.go
```diff
@@ -23,19 +23,32 @@ func Host() string {
...
+		}
+		if url.Scheme == "https" {
+			url.Host = fmt.Sprint(url.Host, ":443")
+		}
 	}
-	if url.Scheme == "" {
-		url.Scheme = "http"
+
+	//maintain backwards compatibility with first version FN_API_URL env vars
+	if url.Path == "" {
```
rdallman: `|| url.Path == "/"` might be nice, too?

PR #252, opened 2018-04-30
Fix issue where configs are actully merged into functions. 

### LINKED ISSUE #250: Config mapping from app.yml file goes into route config instead of an app config
## problem

when you would try to deploy an application using `app.yml` file, the configuration mapping (if present) will not go to an app config but will be merged into a route config.

## steps to reproduce

1. Create `app.yml`:
```yaml
name: test-app
config:
    A: "a"
    B: "b"
```
2. Create a function:
```bash
fn init --runtime go test-function
```
3. Deploy:
```bash
fn deploy --all --registry `whoami`
```
4. Inspect the app:
```json
{
	"created_at": "2018-04-27T00:48:37.639+03:00",
	"name": "test-app",
	"updated_at": "2018-04-27T00:48:37.639+03:00"
}
```
5. Inspect a route:
```json
{
	"config": {
		"A": "a",
		"B": "b"
	},
	"created_at": "2018-04-27T00:48:37.643+03:00",
	"format": "json",
	"idle_timeout": 200,
	"image": "...",
	"memory": 1024,
	"path": "...",
	"timeout": 60,
	"type": "sync",
	"updated_at": "2018-04-27T00:48:37.643+03:00"
}
```

## expected behaviour

If config is present in the app file it should end up in an application config, but not be merged into a route config.

hibooboo2:
🐛Closes #250 Actually patch the app config if there is an app file so that the functions get the apps config.

### DIFF
--- a/deploy.go
+++ b/deploy.go
@@ -153,22 +153,30 @@ func (p *deploycmd) deploySingle(c *cli.Context, appName string, appf *appfile) 
 		}
 	}
 
-	err = p.deployFunc(c, appName, wd, fpath, ff)
-	if err != nil {
-		return err
-	}
-
 	if appf != nil {
 		err = p.updateAppConfig(appf)
 		if err != nil {
 			return fmt.Errorf("failed to update app config: %v", err)
 		}
 	}
+
+	err = p.deployFunc(c, appName, wd, fpath, ff)
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
 // deployAll deploys all functions in an app.
 func (p *deploycmd) deployAll(c *cli.Context, appName string, appf *appfile) error {
+	if appf != nil {
+		err := p.updateAppConfig(appf)
+		if err != nil {
+			return fmt.Errorf("failed to update app config: %v", err)
+		}
+	}
+
 	wd := getWd()
 
 	var funcFound bool
@@ -214,13 +222,6 @@ func (p *deploycmd) deployAll(c *cli.Context, appName string, appf *appfile) err
 
 	if !funcFound {
 		return errors.New("no functions found to deploy")
-	}
-
-	if appf != nil {
-		err := p.updateAppConfig(appf)
-		if err != nil {
-			return fmt.Errorf("failed to update app config: %v", err)
-		}
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


deploy.go
```diff
@@ -312,3 +331,19 @@ func isstale(path string) bool {
 
 	return err != nil
 }
+
+func (p *deploycmd) updateAppConfig(appf *appfile) error {
+	log.Println(appf.Name, "Hi james")
+	log.Println(appf.Config)
```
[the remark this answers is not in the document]
hibooboo2: Ohh good catch

hibooboo2:
@rdallman If we do config before the deploy then that means we have to create the app if it does not exist. That logic should still be left to the deploy not the app. That way if the deploy fails for what ever reason the config is not re updated... Not sure if we should even have this work in this way. Its hard to say since I do not think we have that mapped out. What ever we want to do is fine. The way I did it felt like the least invasive / most in line with current behavior. 



So I guess real question becomes what  should happen if
- User calls deploy with app file
- App does not exist. Create app and add config before calling deploy?

rdallman:
>  If we do config before the deploy then that means we have to create the app if it does not exist.

ah, good catch. isn't that just a matter of changing PATCH to PUT though? (behavior is same atm, at least) then it will create or update.

it's a bit pedantic I guess, i do think it seems a little more natural to have the reverse happen (we basically have to pick one) where if the update config fails, nothing gets deployed, or if deploys fail after there's just a config change then there's just a config change left laying around vs having deployed functions that may need some updated config [was my thinking].

deploy.go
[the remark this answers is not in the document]
carimura: hah

hibooboo2:
Right so right now it seems that the second option `config change left laying around ` is what happens and if update config fails the nothing is deployed. So seems that you are good with that?
@rdallman

deploy.go
```diff
@@ -341,5 +342,19 @@ func (p *deploycmd) updateAppConfig(appf *appfile) error {
 	}
 
 	_, err := p.Apps.PatchAppsApp(param)
-	return err
+	if err != nil {
+		params := clientApps.NewPostAppsParams()
```
rdallman: so i learned yesterday we don't have a `PUT` for apps. so i will `PUT` my foot in my mouth. the 2.0 spec has PUT instead of POST everywhere (for this and other reasons) -- so anyway, ignore me for now, we don't need to go add PUT just for this, but would you mind leaving a note about ordering here and changing it after we do have PUT?

hibooboo2: Sure

hibooboo2:
@rdallman Updated

deploy.go
```diff
@@ -312,3 +331,30 @@ func isstale(path string) bool {
...
+	param.App = appf.Name
+	param.Body = &models.AppWrapper{
+		App: &models.App{
+			Config: appf.Config,
+		},
+	}
+
+	_, err := p.Apps.PatchAppsApp(param)
+	if err != nil {
+		params := clientApps.NewPostAppsParams() //XXX switch to put when v2.0 Fn
```
rdallman: this works too. clarification of what I meant I guess was that you could switch it back to `deploy->update`, ordering issue aside, even though creating app isn't idempotent in deploy, either. in this case, ordering doesn't matter so much since there are bugs any way we cut it, it's up to you at this point

PR #255, opened 2018-05-01
Allow specifying custom port for fn start

### LINKED ISSUE #247: As a developer I want to be able to start an Fn Server on a port other than 8080 using "fn start" so that I don't have to resort to using "docker run"
This is a simple proposal of just adding a `--port -p` option to `fn start`, e.g.

`fn start --port 8081`

This (or lack of) came up several times during the recent Fn Scandinavia Roadshow - there are typically a couple of reasons that users may want to do this:

1) They already have existing services/processes running on port 8080.
2) They want to launch a second Fn Server process on the same box.

We could take this opportunity to add a symmetric `fn stop [--port]` as well, which would allow users to choose which server to stop when they have started more than one. The CLI Command Structure proposal (#216) includes such an extension if not done here.

riconnon:
This resolves #247

### DIFF
--- a/start.go
+++ b/start.go
@@ -29,6 +29,11 @@ func startCmd() cli.Command {
 				Name:  "env-file",
 				Usage: "Path to Fn server configuration file.",
 			},
+			cli.IntFlag{
+				Name:  "port, p",
+				Value: 8080,
+				Usage: "Specify port number to bind to on the host.",
+			},
 		},
 	}
 }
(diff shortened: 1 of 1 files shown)


riconnon:
We could easily extend this to support multiple instances. With the `contexts` support in the CLI we could also allow specifying a context name to `fn start` and start containers with the context name in the name. We could also then populate a context configuration file for each started fn server.

PR #257, opened 2018-05-02
send correct body to fn deploy --all

### LINKED ISSUE #256: Fn app deploy --all broken in 0.4.82
running through the tutorial https://github.com/fnproject/tutorials/blob/master/Flow102/README.md:  
with 0.4.82
```
$ fn version
Client version:  0.4.82
Server version:  0.3.427

$ cat app.yaml                                                                                                                                                                                                                                                                
name: flow102

$ fn  apps list 
no apps found

$ fn deploy --local --all 

ERROR: failed to update app config: &{<nil>} (*models.Error) is not supported by the TextConsumer, can be resolved by supporting TextUnmarshaler interface

$ fn app create flow102

$ fn apps delploy --local --all 
Deploying grep to app: flow102 at path: /grep
Bumped to version 0.0.12
Building image grep:0.0.12
....

```

with 0.4.80 : 

```
$  fn apps list                                                                                                                                                                                                                                                                no apps found

$ fn version                                                                                                                                                                                                                                                                  Client version:  0.4.80
Server version:  0.3.427

$  fn apps list                                                                                                                                                                                                                                                                no apps found

$ fn deploy --local --all                                                                                                                                                                                                                                                     Deploying grep to app: flow102 at path: /grep
Bumped to version 0.0.13
Building image grep:0.0.13
Updating route /grep using image grep:0.0.13...
Deploying head to app: flow102 at path: /head
Bumped to version 0.0.13
Building image head:0.0.13
Updating route /head using image head:0.0.13...
```

zootalures:
closes #256

### DIFF
--- a/deploy.go
+++ b/deploy.go
@@ -343,15 +343,15 @@ func (p *deploycmd) updateAppConfig(appf *appfile) error {
 
 	_, err := p.Apps.PatchAppsApp(param)
 	if err != nil {
-		params := clientApps.NewPostAppsParams() //XXX switch to put when v2.0 Fn
-		param.Body = &models.AppWrapper{
+		postParams := clientApps.NewPostAppsParams() //XXX switch to put when v2.0 Fn
+		postParams.Body = &models.AppWrapper{
 			App: &models.App{
 				Name:   appf.Name,
 				Config: appf.Config,
 			},
 		}
 
-		_, err = p.Apps.PostApps(params)
+		_, err = p.Apps.PostApps(postParams)
 		if err != nil {
 			return err
 		}


deploy.go
```diff
@@ -343,15 +343,15 @@ func (p *deploycmd) updateAppConfig(appf *appfile) error {
 
 	_, err := p.Apps.PatchAppsApp(param)
 	if err != nil {
-		params := clientApps.NewPostAppsParams() //XXX switch to put when v2.0 Fn
-		param.Body = &models.AppWrapper{
+		postParams := clientApps.NewPostAppsParams() //XXX switch to put when v2.0 Fn
```
denismakogon: I see that only var name changes, so, how this solves the problem? Am i missing something?

zootalures: The original bug is that it’s assigning the body to the function param not the new var, I renamed the var to dismbiguate ‘param’ and ‘params’

denismakogon: okay, can we add a test for this?

zootalures: Not easily right now, will look at improving CLI tests afterwards

PR #259, opened 2018-05-02
[258] Support build-arg in fn build, run, deploy.

tzezula:
We have a function with a custom `pom.xml` and custom` Dockerfile`. The build has multiple profiles (default, release). The release build contains more optimisations. To support this we need to pass build arguments to the `Docker build` command. Currently there is no support for this in the `fn run` or in `fn deploy`.

This Pull Request adds a possibility to pass the `build-arg` to `fn build`, `fn run` and `fn deploy`.

Open questions:
1. The `Dockerfile` has to have an `ARG` directive for each `build-arg`. For our case it's not a problem as we have a custom `Dockerfile`. How to support the `build-arg` with `LanguageHelper`s?

### DIFF
--- a/common.go
+++ b/common.go
@@ -36,7 +36,7 @@ func getWd() string {
 	return wd
 }
 
-func buildfunc(c *cli.Context, fpath string, funcfile *funcfile, noCache bool) (*funcfile, error) {
+func buildfunc(c *cli.Context, fpath string, funcfile *funcfile, buildArg []string, noCache bool) (*funcfile, error) {
 	var err error
 	if funcfile.Version == "" {
 		funcfile, err = bumpIt(fpath, Patch)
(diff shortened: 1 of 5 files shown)


tzezula:
Yes, I am Oracle employee.

rdallman:
d'oh, great. cla bot didn't seem to pick up public @oracle stamp, hmm. big green button is blocked here, will attempt to merge this manually. thanks

common.go
```diff
@@ -153,6 +153,12 @@ func runBuild(c *cli.Context, dir, imageName, dockerfile string, noCache bool) e
 		if noCache {
 			args = append(args, "--no-cache")
 		}
+
+		if buildArgs != nil && len(buildArgs) > 0 {
```
zootalures: FWIW the nil check isn't necessary here as an empty slice has len(0)

zootalures:
+1 on this - There are bunch of things we can make configurable in (e.g. the Java build) with  this after the fact

common.go
tzezula: Thanks!
I will fix.

rdallman:
thank you!

PR #261, opened 2018-05-02
Update context command to edit current context files. 

JadeRedworth:
The sectoin of issue: [#217](https://github.com/fnproject/cli/issues/217)

The update command reads in the `current-context` file and identifies if the <key> provided by the user is valid within the file. If so updates the <value>, if not a new key/value pair is added.

Added validation for `api-url` when creating and updating and added tests for this validation.

Also some general code tidy up/refactoring within the context command file.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 
 	"github.com/fnproject/cli/utils"
-
 	"github.com/spf13/viper"
 	"github.com/urfave/cli"
 )
--- a/config/config.go
+++ b/config/config.go
@@ -2,16 +2,14 @@ package config
 
 import (
 	"fmt"
-	"io/ioutil"
-	"log"
 	"os"
 	"path/filepath"
 	"strings"
 
-	homedir "github.com/mitchellh/go-homedir"
+	"github.com/fnproject/cli/utils"
+
 	"github.com/spf13/viper"
 	"github.com/urfave/cli"
-	yaml "gopkg.in/yaml.v2"
 )
 
 const (
@@ -24,7 +22,7 @@ const (
 	defaultLocalAPIURL     = "http://localhost:8080/v1"
 	DefaultProvider        = "default"
 
-	readWritePerms = os.FileMode(0755)
+	ReadWritePerms = os.FileMode(0755)
 
 	CurrentContext  = "current-context"
 	ContextProvider = "provider"
@@ -40,8 +38,8 @@ const (
 	OracleDisableCerts  = "disable-certs"
 )
 
-var defaultRootConfigContents = map[string]string{CurrentContext: ""}
-var defaultContextConfigContents = map[string]string{
+var defaultRootConfigContents = &utils.ContextMap{CurrentContext: ""}
+var defaultContextConfigContents = &utils.ContextMap{
 	ContextProvider: DefaultProvider,
 	EnvFnAPIURL:     defaultLocalAPIURL,
 	EnvFnRegistry:   "",
@@ -70,23 +68,23 @@ func Init() error {
 // EnsureConfiguration ensures context configuration directory hierarchy is in place, if not
 // creates it and the default context configuration files
 func ensureConfiguration() error {
-	home := GetHomeDir()
+	home := utils.GetHomeDir()
 
 	rootConfigPath := filepath.Join(home, rootConfigPathName)
 	if _, err := os.Stat(rootConfigPath); os.IsNotExist(err) {
-		if err = os.Mkdir(rootConfigPath, readWritePerms); err != nil {
+		if err = os.Mkdir(rootConfigPath, ReadWritePerms); err != nil {
 			return fmt.Errorf("error creating .fn directory %v", err)
 		}
 	}
 
 	contextConfigFilePath := filepath.Join(rootConfigPath, contextConfigFileName)
 	if _, err := os.Stat(contextConfigFilePath); os.IsNotExist(err) {
-		_, err = os.Create(contextConfigFilePath)
+		file, err := os.Create(contextConfigFilePath)
 		if err != nil {
 			return fmt.Errorf("error creating config.yaml file %v", err)
 		}
 
-		err = WriteYamlFile(contextConfigFilePath, defaultRootConfigContents)
+		err = utils.WriteYamlFile(file, defaultRootConfigContents)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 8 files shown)


context.go
```diff
@@ -258,6 +274,43 @@ func getAvailableContexts() ([]os.FileInfo, error) {
 	return files, nil
 }
 
+func updateContextFile(file map[string]string, key string, value string) (map[string]string, error) {
+	contextValues := map[string]string{}
+	var validContextKey bool
+
+	for k, v := range file {
```
denismakogon: I feel this function is overcomplicated. As far as I can see this function updates the context with a new key-value pair.
So, according to the code `contextValue` var is literally a copy of `file` var except 2 checks:
 - value by key already exists
 - key is Fn API URL.

all you need to do is:
 - take `file`
 - see if value exists `value, ok := file[key]`
 - if yes - return an error
 - if no - check if key is Fn API URL, if so - validate
 - set kev-value

basically, you don't have to loop through the map here every time you need to update the context

Instead of creating multiple instances of `map[string]string` you better be declaring a type:
```go
type ContextMap map[string]string
```
and declare few methods for it, like:
```
func (ctxMap *ContextMap) Set(key, value string) error
func (ctxMap *ContextMap) Get(key string) string
```
This will give you better readibility and better UX while using those during CLI development if needed.

config/config.go
```diff
@@ -194,7 +194,7 @@ func WriteCurrentContextToConfigFile(value string) error {
 	return nil
 }
 
-func decodeYAMLFile(path string) (map[string]string, error) {
+func DecodeYAMLFile(path string) (map[string]string, error) {
```
denismakogon: instead of making this function exportable in this particular package would it makes sense to put that somewhere in utils?

context.go
```diff
@@ -258,6 +274,43 @@ func getAvailableContexts() ([]os.FileInfo, error) {
...
+	if !validContextKey {
+		contextValues[key] = value
+	}
+
+	fmt.Printf("Successfully updated %v: %v \n", key, value)
+	return contextValues, nil
+}
+
+func ValidateAPIURL(apiURL string) error {
+	_, err := url.ParseRequestURI(apiURL)
```
denismakogon: I'm not sure if that is a valid method to validate the Fn API URL endpoint,  a good starting point is: https://github.com/fnproject/cli/blob/master/client/api.go#L30 see what we check there and try to make a validation rule out of that logic.

config/config.go
```diff
@@ -194,25 +180,26 @@ func WriteCurrentContextToConfigFile(value string) error {
...
 	if err != nil {
-		return nil, err
+		return err
 	}
-	return yf, err
-}
 
-func GetHomeDir() string {
-	home, err := homedir.Dir()
+	err = ioutil.WriteFile(filename, marshaled, ReadWritePerms)
```
denismakogon: this would be enough, no need to double-check the error, if it's there, just return that from `ioutil.WriteFile`:
```go
return ioutil.WriteFile(filename, marshaled, ReadWritePerms)
```

config/config.go
```diff
@@ -152,26 +153,11 @@ func LoadConfiguration(c *cli.Context) error {
...
-
-	return nil
-}
-
 func WriteCurrentContextToConfigFile(value string) error {
 	home := GetHomeDir()
 
 	configFilePath := filepath.Join(home, rootConfigPathName, contextConfigFileName)
-	file, err := DecodeYAMLFile(configFilePath)
+	file, err := utils.DecodeYAMLFile(configFilePath)
```
denismakogon:
```go
file, err := utils.DecodeYAMLFile(configFilePath)
return err
```

context.go
```diff
@@ -319,3 +290,21 @@ func ValidateContextName(context string) error {
 	}
 	return nil
 }
+
+func (ctxMap *ContextMap) Set(key, value string) error {
+	contextFilePath := createFilePath(viper.GetString(config.CurrentContext) + ".yaml")
+	file, err := utils.DecodeYAMLFile(contextFilePath)
```
denismakogon: <- here you open a file to read its content.

context.go
```diff
@@ -319,3 +290,21 @@ func ValidateContextName(context string) error {
...
+
+	if key == config.EnvFnAPIURL {
+		err := ValidateAPIURL(value)
+		if err != nil {
+			return err
+		}
+	}
+
+	file[key] = value
+	return config.WriteYamlFile(contextFilePath, file)
```
denismakogon: <- here you would open a file one more time to write back a new content, so what i recommend it to change the signature of these two calls to accept `os.File` to do less redundant IO operations here. But it's up to you.

utils/utils.go
```diff
@@ -0,0 +1,42 @@
...
+	"os"
+
+	yaml "gopkg.in/yaml.v2"
+)
+
+const (
+	readWritePerms = os.FileMode(0755)
+)
+
+func DecodeYAMLFile(path string) (map[string]string, error) {
```
denismakogon:
```go
func DecodeYAMLFile(path string) (*ContextMap, error) {
      f, err := os.Open(path)
      if err != nil {
            return err
      }
      var yf ContextMap
      err = yaml.NewDecoder(f).Decode(&yf)
      return &yf, err
}
```
or use `os.OpenFile`

utils/utils.go
```diff
@@ -0,0 +1,42 @@
...
+	yf := map[string]string{}
+	err = yaml.Unmarshal(b, yf)
+	if err != nil {
+		return nil, err
+	}
+	return yf, err
+}
+
+// WriteYamlFile writes to the yaml file
+func WriteYamlFile(filename string, value map[string]string) error {
```
denismakogon:
```go
func WriteYamlFile(path string, value *ContextMap) error {
      f, err := os.Open(path)
      if err != nil {
            return err
      }
      return yaml.NewEncoder(f).Encode(value)
}
```
or use `os.OpenFile`

context.go
```diff
@@ -274,39 +270,14 @@ func getAvailableContexts() ([]os.FileInfo, error) {
...
-	fmt.Printf("Successfully updated %v: %v \n", key, value)
-	return contextValues, nil
-}
-
-func ValidateAPIURL(apiURL string) error {
-	_, err := url.ParseRequestURI(apiURL)
+	_, err := url.Parse(apiURL)
 	if err != nil {
-		return errors.New("please enter a valid api-url")
+		return fmt.Errorf("invalid %s", err)
```
denismakogon: you mean `invalid Fn API URL: %s` ?

config/config.go
denismakogon: also, would you mind to change var name from `file` to `ctxMap`?

context.go
```diff
@@ -229,18 +233,26 @@ func list(c *cli.Context) error {
...
+	path := filepath.Join(home, contextsPath, filename)
+	return path
 }
 
 func checkContextFileExists(filename string) (bool, error) {
-	path, err := createFilePath(filename)
-	if err != nil {
-		return false, err
-	}
+	path := createFilePath(filename + ".yaml")
```
denismakogon: so, we enforce to have `*.yaml` only, what about `*.yml`? just out of curiour

utils/utils.go
```diff
@@ -0,0 +1,53 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func GetHomeDir() string {
+	home, err := homedir.Dir()
+	if err != nil {
+		log.Fatalln("could not get home directory:", err)
```
denismakogon:
```go
log.Fatalf("could not get home directory: %s\n", err)
```

Issue #264, opened 2018-05-04
For FDKs which uses blocking request stream readers CLI should abort container based on number of runs


denismakogon:
Here's the thing. FDK-Python has its weak point - stream reader. Python doesn't have non-blocking readers, every time it calls `.read()` with any stream, it blocks until the data is passed in that stream.
I've made a PR for FDK https://github.com/fnproject/fdk-python/pull/40 to make stream reader nicer to `fn run` and `fn test`, however, as soon as FDK spits out the response the stream it used to read the request blocks forever.

So, since Fn CLI uses docker CLI to manage containers, it should abort a container execution once it container provided protocol frame, of course, with respect to `--runs`, in order to avoid container hang forever.

Issue #156, opened 2018-02-06
use same code as fn to execute docker containers


rdallman:
right now we're using the docker cli, we should probably use the agent code from fn but it becomes something of an import issue. we used to extract just the driver and we could do that again, but there is a lot of machinery around hot functions that may need scoping to see what we need from there. since `fn run` really is only one invocation, we probably mostly just need the driver and possibly the dispatch code, maybe some other bits.

https://github.com/fnproject/fn/issues/241

denismakogon:
related to https://github.com/fnproject/fdk-python/issues/36, for some reason, CLI keep sending the `\n` between the requests, but the agent doesn't. So it makes FDK-python for the local mode of the `fn test` and `fn run`.

zootalures:
Closing as we don't do fn run any more

PR #265, opened 2018-05-04
Make "run" great again

denismakogon:
This is a temporary fix until we get #156 done
Closes: #264

### DIFF
--- a/run.go
+++ b/run.go
@@ -151,8 +151,15 @@ func (r *runCmd) run(c *cli.Context) error {
 
 // TODO: share all this stuff with the Docker driver in server or better yet, actually use the Docker driver
 func runff(ff *funcfile, stdin io.Reader, stdout, stderr io.Writer, method string, envVars []string, links []string, format string, runs int) error {
-	sh := []string{"docker", "run", "--rm", "-i", fmt.Sprintf("--memory=%dm", ff.Memory)}
-
+	startFunc := []string{
+		"docker", "run",
+		fmt.Sprintf("--name=%s", ff.Name),
+		"--rm", "-i",
+		fmt.Sprintf("--memory=%dm", ff.Memory),
+	}
+	stopFunc := []string{
+		"docker", "rm", "-f", ff.Name,
+	}
 	var env []string    // env for the shelled out docker run command
 	var runEnv []string // env to pass into the container via -e's
 	callID := "12345678901234567890123456"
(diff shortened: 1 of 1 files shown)


run.go
```diff
@@ -151,8 +151,15 @@ func (r *runCmd) run(c *cli.Context) error {
 
 // TODO: share all this stuff with the Docker driver in server or better yet, actually use the Docker driver
 func runff(ff *funcfile, stdin io.Reader, stdout, stderr io.Writer, method string, envVars []string, links []string, format string, runs int) error {
-	sh := []string{"docker", "run", "--rm", "-i", fmt.Sprintf("--memory=%dm", ff.Memory)}
-
+	startFunc := []string{
+		"docker", "run",
+		fmt.Sprintf("--name=%s", ff.Name),
```
zootalures: Won't this limit you to one running instance of a function at a time?

denismakogon: it will, posting new patch to use hash to get a unique name here

zootalures:
I don't get why this is necessary - 

For a hot function the FDK should terminate the process when the input stream is closed  - Docker run with input should close stdin which should cause the FDK to exist.

denismakogon:
That's the point, for instance, Docker doesn't close the stream, it continuously posts `\n` and Python thinks that stream is still open and hangs still hangs.

zootalures:
That's the point, for instance, Docker doesn't close the stream, it continuously posts \n and Python thinks that stream is still open and hangs still hangs.

that is not my understanding of what happens:   

e.g. in my go hot function  below (with http at least)  I see an EOF when the function data is passed: 

```
OCliffe@OCLIFFE-mac /p/t/gohot> cat func.yaml 
name: gohot
version: 0.0.1
runtime: go
entrypoint: ./func
format: http
OCliffe@OCLIFFE-mac /p/t/gohot> cat func.go 
package main

import (
        "os"
        "log"
)

func main() {

        buf := make([]byte, 100)
        log.Printf("Got format %s",os.Getenv("FN_FORMAT"))
        for {
                read, err := os.Stdin.Read(buf)
                if err != nil {
                        log.Printf("Got error %v", err)
                        os.Exit(0)
                }
                log.Printf("Read %d", read)

        }

}
OCliffe@OCLIFFE-mac /p/t/gohot> fn version   
Client version:  0.4.84
Server version:  ?
OCliffe@OCLIFFE-mac /p/t/gohot> fn run       
Building image gohot:0.0.1 ..
2018/05/04 12:07:18 Got format http
2018/05/04 12:07:18 Read 100
2018/05/04 12:07:18 Read 100
2018/05/04 12:07:18 Read 62
2018/05/04 12:07:18 Got error EOF
OCliffe@OCLIFFE-mac /p/t/gohot> 
```

PR #266, opened 2018-05-04
Improve python builds

denismakogon:
- shrink image size up to 3 times
 - get rid of pip caching

### DIFF
--- a/langs/python.go
+++ b/langs/python.go
@@ -103,7 +103,7 @@ def handler(ctx, data=None, loop=None):
     if data and len(data) > 0:
         body = json.loads(data)
         name = body.get("name")
-    return "Hello {0}".format(name)
+    return {"message": "Hello {0}".format(name)}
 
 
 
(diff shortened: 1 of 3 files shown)


langs/python.go
```diff
@@ -55,19 +55,19 @@ func (h *PythonLangHelper) Runtime() string {
...
 }
 
 func (h *PythonLangHelper) BuildFromImage() (string, error) {
-	return fmt.Sprintf("python:%v", h.Version), nil
+	return fmt.Sprintf("python:%s-slim-stretch", h.Version), nil
 }
 
 func (h *PythonLangHelper) RunFromImage() (string, error) {
-	return fmt.Sprintf("fnproject/python:%v", h.Version), nil
+	return fmt.Sprintf("python:%s-slim-stretch", h.Version), nil
```
zootalures: I think this should stay being an image that we curate.

denismakogon: There's no reason to do so, since `python:3.6.X-slim-stretch` is a public image from Docker library (not from someone's namespace) that needs exactly 2 packages.

BTW, this image is not used, because python doesn't use multi-stage builds, so the only image used - `python:3.6` which we do not control.

zootalures: Fine let's come back to the image thing  later - otheriwse this LGTM - good stuff!

PR #277, opened 2018-05-10
Read Oracle credentials from /.oci/config file

JadeRedworth:
Replace `key-id` in context file with the 3 join values, `oracle.tenancy-id`, `oracle.user-id` & `oracle.fingerprint` and join them together in code instead.

Replace panics with prints to stderr and os.Exit() if pass-phrase is wrong.

Try to get private Key without a password in the case the private key isn't encrypted. Check if pass-phrase is present in the context file if so use don't prompt each time for a password.

Read from `/.oci/config` if values are not present in the context file. 

Add `profile` key to Oracle provider context files which enables users to use the different profiles present in the `config` file. If no provider is specified, default to "DEFAULT"

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -1,22 +1,16 @@
 package client
 
 import (
-	"crypto/rsa"
 	"fmt"
-	"io/ioutil"
 	"net/url"
 	"os"
-	"path/filepath"
 	"strings"
-	"syscall"
 
 	"github.com/fnproject/cli/config"
 	fnclient "github.com/fnproject/fn_go/client"
 	openapi "github.com/go-openapi/runtime/client"
 	"github.com/go-openapi/strfmt"
-	oci "github.com/oracle/oci-go-sdk/common"
 	"github.com/spf13/viper"
-	"golang.org/x/crypto/ssh/terminal"
 )
 
 func Host() string {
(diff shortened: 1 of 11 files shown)


msgodf:
This is needed for `fn call` to invoke a function with data, as that works by reading stdin, and currently gives this error:

```
Private Key Phrase: panic: inappropriate ioctl for device

goroutine 1 [running]:
github.com/fnproject/cli/client.challengeForPKeyPassword(0xc420025960, 0x1a)
...
```

client/api.go
```diff
@@ -74,23 +74,42 @@ func challengeForPKeyPassword() string {
...
+	return key, nil
+}
+
+func getPrivateKey(keyBytes []byte, pKeyPword, pkeyFilePath string) *rsa.PrivateKey {
+	key, err := oci.PrivateKeyFromBytes(keyBytes, oci.String(pKeyPword))
 	if err != nil {
-		panic(fmt.Sprintf("Unable to load private key from file bytes: %s. Error: %s", pkeyFilePath, err))
+		if pKeyPword != "" {
+			fmt.Fprintf(os.Stderr, "Unable to load private key from file bytes: %s. Error: %s \n", pkeyFilePath, err)
+			os.Exit(1)
```
denismakogon: `log.Fatal` would be better

zootalures: Propagating the error up would be even better :)

client/api.go
```diff
@@ -74,23 +75,41 @@ func challengeForPKeyPassword() string {
 	return password
 }
 
-func privateKey(pkeyFilePath string) *rsa.PrivateKey {
+func privateKey(pkeyFilePath string) (*rsa.PrivateKey, error) {
```
zootalures: Feels like this is oracle specific - could it be moved int oracle_provider.go?

client/api.go
```diff
@@ -104,24 +123,84 @@ func oracleProvider(transport *openapi.Runtime) {
 				keyID,
 				pKey,
 				transport.Transport))
+
+	return err
+}
+
+func oracleConfigFile() (string, *rsa.PrivateKey, error) {
```
zootalures: feels like this belongs in oracle_provider.go

config/config.go
```diff
@@ -30,10 +30,14 @@ const (
 	EnvFnAPIURL   = "api-url"
 	EnvFnContext  = "context"
 
-	OracleKeyID         = "key-id"
-	OraclePrivateKey    = "private-key"
-	OracleCompartmentID = "compartment-id"
-	OracleDisableCerts  = "disable-certs"
+	OracleTenancyID     = "oracle.tenancy-id"
```
zootalures: Can these constants all be moved under oracle_provider now?

client/api.go
JadeRedworth: Sure!

zootalures:
Should error in the CLI if no compartment ID is set

JadeRedworth:
@zootalures @denismakogon made the changes you both suggested, could I get a final review so we can get this in?

PR #282, opened 2018-05-14
Support setting annotations on apps and routes

msgodf:
This allows annotations to be set when creating or updating apps and routes.

As an example, an annotation can be added to a route when creating it:

```
fn routes create --annotation key='value' appname /routename
```

Where `'value'` should be a string containing valid JSON.

### DIFF
--- a/routes.go
+++ b/routes.go
@@ -57,8 +57,8 @@ var routeFlags = []cli.Flag{
 		Usage: "route idle timeout (eg. 30)",
 	},
 	cli.StringSliceFlag{
-		Name:  "annotations",
-		Usage: "route annotations",
+		Name:  "annotation",
+		Usage: "route annotation (can be specified multiple times)",
 	},
 }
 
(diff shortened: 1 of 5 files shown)


routes.go
```diff
@@ -311,6 +315,26 @@ func routeWithFlags(c *cli.Context, rt *fnmodels.Route) {
 			rt.Config = extractEnvConfig(c.StringSlice("config"))
 		}
 	}
+	if len(rt.Annotations) == 0 {
```
denismakogon: at first look, it seems like a code duplication, no?

msgodf: Between this and the equivalent in apps.go?

denismakogon: Yes.

msgodf: I agree. What suitable shared abstraction and package would you suggest? It seems odd to favour one or the other, and reducing locality by putting into the common package seemed unnecessary.

denismakogon: i guess new function can find its home here: https://github.com/fnproject/cli/blob/master/common.go

msgodf: Great, thanks for the suggestion. I've pushed a change with the extracted function.

zootalures:
Can you rebase so I can test properly with the latest changes?

msgodf:
@zootalures I have now (I believe)

denismakogon:
LGTM, however, i'd like to see some tests for this, is it possible?

msgodf:
@denismakogon for the use of the flag itself? I can't see any precedent for this in the CLI tests. If you know of some already and I have overlooked them then please point me in the right direction.

denismakogon:
I guess you need to create a test suite here: https://github.com/fnproject/cli/blob/master/test/cli_crud_test.go

msgodf:
@zootalures I extracted the JSON query functionality from the CLI integration tests into the harness - if you get time could you take a quick look to make sure that isn't against the spirit or intention of how you wrote it? Thanks!

msgodf:
@denismakogon I've added tests for the `--annotation` flag now, that test the behaviour when the flag is used correctly, when the value is invalid, and how annotations can be removed.

PR #288, opened 2018-05-22
Add Oracle provider authentication to `fn call`

msgodf:
This uses a similar `switch` to that in `client/api.go` to check the configured provider and use that to set up the appropriate `RoundTripper` implementation to make the call. Some code in `client/api.go` that sets up the `RoundTripper` for the `openapi` client is made more generic so that it can be reused. This could eventually be abstracted into a method implemented by any configured provider.

Additionally, this allows `fn call` to work for HTTPS, as it reuses the `HostURL` function from `client/api.go`, rather than the URL being constructed in `routes.go`.

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"fmt"
+	"net/http"
 	"net/url"
 	"os"
 	"strings"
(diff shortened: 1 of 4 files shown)


zootalures:
works - but it's a PITA having to change the API url between  API operations and Calls.

msgodf:
@zootalures I have just the fix - separate PR to follow, based on these changes.

msgodf:
@zootalures it's in https://github.com/fnproject/cli/pull/289

PR #289, opened 2018-05-22
Allow (optional) separate URL for `fn call`

msgodf:
Support an optional separate endpoint for fn call.

This is configured with either the context option `call-url`, or the environment variable `FN_CALL_URL`.

If unset, then this defaults to the existing choice, which is to use the value of the context option `api-url`, or the environment variable `FN_API_URL`.


msgodf:
I've reapplied onto the `fn call` changes on master.

PR #290, opened 2018-05-23
Route annotations can be specified in func.yaml.

msgodf:
Annotations for a route can be specified using `route_annotations` in
the func.yaml. For example:

```
route_annotations:
    a: x
```

Here, `x` can be any valid JSON or YAML value. This allows more
complex annotation values to be specified, such as arrays:

```
route_annotations:
    a:
    - x
    - y
```

Or maps:

```
route_annotations:
    a:
        b: 1
        c: 2
```

### DIFF
--- a/funcfile.go
+++ b/funcfile.go
@@ -60,15 +60,16 @@ type funcfile struct {
 
 	// Route params
 	// TODO embed models.Route
-	Type        string              `yaml:"type,omitempty" json:"type,omitempty"`
-	Memory      uint64              `yaml:"memory,omitempty" json:"memory,omitempty"`
-	Cpus        string              `yaml:"cpus,omitempty" json:"cpus,omitempty"`
-	Format      string              `yaml:"format,omitempty" json:"format,omitempty"`
-	Timeout     *int32              `yaml:"timeout,omitempty" json:"timeout,omitempty"`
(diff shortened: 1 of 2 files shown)


msgodf:
Personally I don't love the use of underscores for the field name (`route_annotations`), but as there was already precedent in `idle_timeout`, I opted for consistency.

zootalures:
is the `route_` stuttering - i.e. isn't everything here implicitly defined on a route ?

denismakogon:
I agree with Owen, it should be `annotation`.

Okay, this PR adds annotation to routes, what about application?

msgodf:
It is! I wanted to leave the option to add app annotations as well though - and couldn't think of a nice way to do that short of prefixing. Nesting would be another option. Or not supporting app annotations.

msgodf:
@denismakogon first point is addressed above, second point - I like small, specific PRs.

denismakogon:
Wa already have `app.yml`, that is exact place to put app annotations there.

msgodf:
@denismakogon I was completely unaware of `app.yaml`. Sure, they can go there then! I'll rename the field.

msgodf:
@zootalures @denismakogon I've renamed the field. Thanks for the clarification and suggestions.

Issue #216, opened 2018-04-04
As an application developer I want a consistent CLI command structure so that I can easily reason about command usage.


JadeRedworth:
This issue changes the Fn CLI to use a [verb][noun] structure, resulting in a more intuitive, human-readable approach that would bring a consistent feel to the CLI. It is a common pattern among other CLI's that developers may be using alongside Fn. (A more detailed analysis of the Fn CLI and other CLI's can be found here: [CLI Analysis](https://github.com/fnproject/cli/wiki/CLI-Analysis))

The [Verb Noun matrix](https://github.com/fnproject/cli/wiki/CLI-Analysis#verb-to-noun-matrix) shows how many of the nouns are unrelated to a verb. Therefore these should be their own high-level commands along with all other verbs:

|verbs|nouns|
|:---:|:---:|
|build|app[s]|
|bump|call[s]|
|call|config|
|create|context|
|delete|lambda|
|deploy|logs|
|get|route[s]|
|help|server|
|import||
|init||
|list||
|pull?||
|run||
|set||
|start||
|stop||
|test||
|update||
|use||

There is confusion about the use of apps/routes vs app/route (singular use of `app/route/call/image`). Looking back at [Singular vs Plural](https://github.com/fnproject/cli/wiki/CLI-Analysis#singular-vs-plural-vs-both) the use of both plural and singular nouns are both valid:
```
~ $ fn create app/apps <app>
~ $ fn delete route/routes <app> </route>
```
When using 'fn get' and 'fn list', get is used to return a single object, specificly a function call, and list is used to return multiple apps/routes/calls etc. The command 'inspect' allows you to retrieve on or all apps/routes properties. However, this essentially performs the same as 'get' and will, therefore, be deprecated.
```
~ $ fn list apps
java-app
string-reverse-app

~ $ fn list routes java-app
/hello-java-function  <dockerhub_username>/hello2:0.0.2	   localhost:8080/r/java-app/hello-java-function
/test-route           <dockerhub_username>/hello2:0.0.2	   localhost:8080/r/java-app/test-route

~ $ fn list calls <app>
ID: 01C9RQPM7G47WG200000000000
App: java-app
Route: /hello-java-function
Created At: 2018-03-29T11:18:49.840Z
Started At: 2018-03-29T11:18:50.366Z
Completed At: 2018-03-29T11:18:50.367Z
Status: success

ID: 01C9RG0N7147WGA00000000000
App: java-app
Route: /test-route  
[...]

Issue #13, opened 2017-08-29
Implement --force for apps delete


denismakogon:
This appears to be annoying very often.
Assume developer have 10+ routes and in order to dismiss app he needs to go through all routes before deleting an app.
So having `--force` flag (default to `False`) will make UX way more better.

treeder:
Closing, should happen by default and be server side.

PR #291, opened 2018-05-24
Command reconstruction - verb noun

JadeRedworth:
Reconstruction of the command structure within the CLI. It has been changed from a `noun` `verb` to `verb` `noun` structure, as described in issue [#216](https://github.com/fnproject/cli/issues/216)

**Folders added**:
- Commands - all the top-level commands
- Test - holds all the test files
- Object
    - app  
    - call
    - context
    - log
    - route
    - server
- Common
- Run - all run related files

**Open for suggestions:** I don’t like the way the structure no handles apps and routes config. 
`fn list config apps` I think it would be nicer to have `fn list apps config` but urfave will be expecting that 4th command and therefore `fn list apps` will not be valid.

^^ We went with `fn config [app | route] <key> <value>` to set a configuration.

**Following changes to documentation and commands**:
Fn [#1031](https://github.com/fnproject/fn/pull/1031)
Tutorials [#87](https://github.com/fnproject/tutorials/pull/87)
Flow Lib Go [#5](https://github.com/fnproject/flow-lib-go/pull/5)
Fdk-ruby [#13](https://github.com/fnproject/fdk-ruby/pull/13)
Fdk-java [#138](https://github.com/fnproject/fdk-java/pull/138)


denismakogon:
Would be nice to get https://github.com/fnproject/cli/issues/296 done as part of this change as well.

denismakogon:
At least one of the tests is broken:
```
2018/05/25 12:41:31 Expected sdterr message (Unable to parse annotation value 'value'. Annotations values must be valid JSON strings.) not found in result: COMMAND: /home/circleci/go/src/github.com/fnproject/cli/fn create routes xpldnjob myroute --image foo/duffimage:0.0.1 --annotation test=value
INPUT:
RESULT: true
```
maybe a regression?

zootalures:
Black box review: 

*  can you accept both plural and singular of of apps and app , call/calls log/logs etc everywhere? 
* add ‘ls’ as an alias for list 
* Remove lambda command ?  or put it at the bottom or hide it  - it doesn't work anyway 
* fn push docs:  “push function to Docker Hub” - > Push function to docker regsitry 


Also `fn set app foo  key value` is a synonym for `fn update app foo -c key=value` 
not sure it botheres me that much but just noticed.

JadeRedworth:
In regard to single vs plural, we can one command name and then multiple aliases. It will just change the help text to have the 3 options `apps`, `app`, `a` etc.. If this is fine then it can be added?

test/routes_test.go
```diff
@@ -0,0 +1,27 @@
+package test
```
zootalures: No unit tests in /test  (sorry should have called it "e2e_test" or something - I think this test already exists in client/call_fn_test.go (my bad)

common/funcfile.go
```diff
@@ -22,17 +22,17 @@ var (
 	}
 )
 
-type inputMap struct {
+type InputMap struct {
```
zootalures: Dox on all the structs in this file now their public.

objects/server/server.go
```diff
@@ -58,7 +64,7 @@ func (b *buildServerCmd) flags() []cli.Flag {
 // * Generate a Dockerfile that gets all the extensions (using dep)
 // • ‎then generate a main.go with extensions
 // • ‎compile, throw in another container like main dockerfile
-func (b *buildServerCmd) buildServer(c *cli.Context) error {
+func (b *buildServerCmd) Build(c *cli.Context) error {
```
zootalures: Don't see this referenced  anywhere 

Suggest you put this as its own top-level command rather than under the build verb 

`fn build-server` ?

zootalures:
I did a quick scan and I think you pick up most uses of the command that need changing in docs with the following grep: 

```
 fgrep -rE "fn +(apps|routes|calls|logs|context|update) "  *  | grep -v \.git
```

commands/build_server.go
```diff
@@ -12,27 +7,30 @@ import (
...
+	cmd := BuildServerCmd{}
 	flags := append([]cli.Flag{}, cmd.flags()...)
 	return cli.Command{
-		Name:   "build-server",
-		Usage:  "build custom fn server",
-		Flags:  flags,
-		Action: cmd.buildServer,
+		Name:     "build-server",
+		Usage:    "build custom fn server",
+		Category: "DEVELOPMENT COMMANDS",
```
zootalures: I think this belongs more closely to `SERVER COMMANDS`

JadeRedworth:
@rikgibson the `help` command comes with urfave so there is no way to add a category.
`version` can be moved to other commands but then it would be on its own?

PR #292, opened 2018-05-24
Missing oci config fix

JadeRedworth:
There is a current bug if the ‘.oci/config’ file is missing. 

I have added a check to ensure it is there before the values are gathered.

### DIFF
--- a/client/oracle_provider.go
+++ b/client/oracle_provider.go
@@ -3,9 +3,11 @@ package client
 import (
 	"crypto/rsa"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"os"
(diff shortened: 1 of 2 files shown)


msgodf:
Worked for me! Just to clarify:  it takes values from `~/.oci/config` *then* the context file?

msgodf:
What I mean by this is an error that says something like

Unable to find tenancyID in configuration: oracle.tenancy-id is missing from current-context file and tenancy-id wasn't found in `~/.oci/config`.

msgodf:
And so on for each missing value...

JadeRedworth:
Yes, the context file should override the config file. I'll change the errors now.

msgodf:
Great, that makes it clearer!

Issue #278, opened 2018-05-10
Fn start puts fn db in CWD, this is confusing 


zootalures:
Fn start is sensitive to its current working directory, as it mounts data in the CWD

I think it should start consistently in any directory (e.g by putting data in ~/.fn/data

PR #295, opened 2018-05-25
Move the creation of fn db to ~/.fn/localdata

JadeRedworth:
This moves the creation of the local data to `.fn/localdata` 
```
- fn 
  - localdata
       - data
            - fn.db
             - fn.mq
```

Related Issue: [#278](https://github.com/fnproject/cli/issues/278)

### DIFF
--- a/start.go
+++ b/start.go
@@ -49,7 +49,7 @@ func start(c *cli.Context) error {
 	home := config.GetHomeDir()
 
 	if c.String("data-dir") != "" {
-		fnDir = filepath.Join(c.String("data-dir"))
+		fnDir = c.String("data-dir")
 	} else {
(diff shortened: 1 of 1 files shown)


start.go
```diff
@@ -34,19 +36,27 @@ func startCmd() cli.Command {
...
 func start(c *cli.Context) error {
-	wd, err := os.Getwd()
-	name := "fnserver"
-	if err != nil {
-		log.Fatalln("Getwd failed:", err)
+	var fnDir string
+	home := config.GetHomeDir()
+
+	if c.String("data-dir") != "" {
+		fnDir = filepath.Join(c.String("data-dir"))
```
zootalures: I think this filepath.Join is redundant.

zootalures:
L.G.T.M.

PR #302, opened 2018-06-04
Read-only root filesystem and tmpfs size support in cli

skinowski:
TODO: TmpFs not yet in fn_go. Will need to add/update fn_go and add TmpFsSize to routes.go.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -18,7 +18,7 @@ export fn="${CUR_DIR}/fn"
 
 
 #on CI these can take a while
-go test -v $(go list ./... |  grep -pv "^github.com/fnproject/cli/test$")
+go test -v $(go list ./... |  grep -v "^github.com/fnproject/cli/test$")
 
 # Our test directory
 OS=$(uname -s)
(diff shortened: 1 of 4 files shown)


skinowski:
fn recently switched to read-only root filesystem by default. To follow up, CLI needs to make read-only root default as well. This PR introduces read-only root filesystem and an option to opt out in 'fn run' command.

test.sh
```diff
@@ -39,21 +39,26 @@ CONTAINER_ID=$($fn start -d)
...
 
-	if [ ${RESULT} -ne 0 ]; then
-		sleep 1
-		TRIES=$((${TRIES}-1))
+	if [ ${RESULT} -eq 0 ]; then
+		break
+	else
+		echo "Going to try ${FN_API_URL} one more time..."
 	fi
-	break
```
skinowski: Bug fix here...

test.sh
```diff
@@ -18,7 +18,7 @@ export fn="${CUR_DIR}/fn"
 
 
 #on CI these can take a while
-go test -v $(go list ./... |  grep -pv "^github.com/fnproject/cli/test$")
```
skinowski: Linux does not like '-p' and sure if follow-sym links is needed here. On mac-os-x, man page of grep says:

"-p      If -R is specified, no symbolic links are followed.  This is the default."

which seems unnecessary in test.sh above.

run.go
```diff
@@ -190,6 +194,16 @@ func runff(ff *funcfile, stdin io.Reader, stdout, stderr io.Writer, method strin
 		}
 	}
 
+	if !rwRoot {
+		sh = append(sh, "--read-only")
+	}
+
+	if ff.TmpFsSize != 0 {
```
skinowski: Let's always try to add a rw tmpfs here. This is also the default behavior in fn agent now.

rdallman:
fn test/run are gone now, gonna close

PR #304, opened 2018-06-06
Update readme with announcement of new command structure

JadeRedworth:


main.go
```diff
@@ -141,6 +141,7 @@ func main() {
 	if err != nil {
 		// TODO: this doesn't seem to get called even when an error returns from a command, but maybe urfave is doing a non zero exit anyways? nope: https://github.com/urfave/cli/issues/610
 		fmt.Fprintf(os.Stderr, "ERROR: %v\n", err)
+		fmt.Fprintf(os.Stderr, "Note: the fn CLI command structure has changed, you may need to update your CLI\n")
```
zootalures: They've already updataed their CLI ! 

They may need to change the command they typed!

JadeRedworth: oh dur, completely miss understood that one.

JadeRedworth: @msgodf they would have to update it to even get this error message, so its just the command.

msgodf: It might be just good to give an example

"If you ran `fn apps list`, try running `fn list apps` instead"

msgodf: @JadeRedworth yes me too :) I misunderstood both that *and* @zootalures's suggestion!

JadeRedworth: This is in the wrong place then as there's an `app.CommandNotFound` func anyway.

main.go
```diff
@@ -68,13 +68,12 @@ LEARN MORE:
 
 	app.CommandNotFound = func(c *cli.Context, cmd string) {
 		fmt.Fprintf(os.Stderr, "Command not found: \"%v\" -- see `fn --help` for more information.\n", cmd)
+		fmt.Fprintf(os.Stderr, "Note: the fn CLI command structure has changed, you may need to update your command\n")
```
msgodf: > update your command

sounds a bit odd - perhaps "change your command to use the new structure"?

Issue #268, opened 2018-05-04
As a developer I would like to run fn cli commands providing a target working directory


shaunsmith:
To better support integration with CI/CD tools and platforms the fn cli should accept a command line option specifying the working directory a given command should execute in.  This support would avoid having to cd to `/a/path/to/my/app` to perform operations.  For example, using current command structure:

`fn build --dir /a/path/to/my/app --all`
`fn run --dir /a/path/to/my/func`
`fn deploy --dir /a/path/to/my/app --all`

Although I used `--dir` in the examples I don't have a strong opinion on the actual command switch name.

JadeRedworth:
We have 2 options here:

1. To specifiy the `--dir` flag only on the commands we want (build, run, deploy..)
OR
2. Create a global flag `--dir` which can be accessed by all commands.

What do you think @shaunsmith ?

shaunsmith:
I think only add `--dir` to commands where it makes sense.  It seems only a few really need it.  I guess I should have provided a list of commands with the enhancement request.  My only defence is that the command structure was going to change so it didn't make sense at the time. 😜

shaunsmith:
@JadeRedworth I took a look a the current fn CLI commands and think `dir` support makes sense for:
- build -- builds the function in `dir`
- bump -- bumps the version in the func.yaml in `dir`
- deploy -- deploys function or app (if `--all` used) in `dir`
- init -- creates a function in `dir`, or in subdirectory of `dir` if function name provided
- run -- runs function in `dir`
- test -- tests function in `dir` using test.json in same folder

Your PR https://github.com/fnproject/cli/pull/309 doesn't handle all of these cases yet.

PR #309, opened 2018-06-07
Add flag option to target a directory to run/build/deploy a function

JadeRedworth:
As a developer I would like to run Fn CLI commands providing a target working directory [#268](https://github.com/fnproject/cli/issues/268)

### DIFF
--- a/commands/init.go
+++ b/commands/init.go
@@ -35,6 +35,7 @@ import (
 
 type initFnCmd struct {
 	force bool
+	wd    string
 	ff    *common.FuncFile
 }
 
@@ -72,8 +73,9 @@ func initFlags(a *initFnCmd) []cli.Flag {
 			Value:       common.InitialVersion,
 		},
 		cli.StringFlag{
-			Name:  "dir",
-			Usage: "specify the working directory to init a function",
+			Name:        "working-dir",
+			Usage:       "specify the working directory to deploy a function, must be the full path.",
+			Destination: &a.wd,
 		},
 	}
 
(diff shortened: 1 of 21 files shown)


commands/init.go
```diff
@@ -71,6 +71,10 @@ func initFlags(a *initFnCmd) []cli.Flag {
 			Destination: &a.ff.Version,
 			Value:       common.InitialVersion,
 		},
+		cli.StringFlag{
+			Name:  "dir",
+			Usage: "specify directory to build a function",
```
rikgibson: the text here should say "init a function" rather than "build"...

common/bump.go
```diff
@@ -63,12 +62,17 @@ func (b *bumpcmd) flags() []cli.Flag {
 			Usage:       "verbose mode",
 			Destination: &b.verbose,
 		},
+		cli.StringFlag{
+			Name:  "dir",
+			Usage: "specify directory to deploy a function",
```
rikgibson: "bump" not "deploy"

shaunsmith:
@rikgibson I agree with the scenario 1-4 semantics you described.  Are you saying the current changes don't align with these semantics or that we need to capture this in help text?

I agree that `dir` is working directory but don't think we need verbose switch names.  `--dir` and `-d` are nice and succinct and the semantics should be covered in the help text.

JadeRedworth:
@shaunsmith These current changes don't align. Rik and I had a catch-up today and I had misinterpreted the use of --dir within the init command. We are on the same page now and I will make the suggested changes.

PR #318, opened 2018-06-26
Update Gopkg.toml manually

JadeRedworth:


msgodf:
Looks good - to clarify, apparently this is the way to update a dep version: https://github.com/golang/dep/blob/master/docs/FAQ.md#how-do-i-change-the-version-of-a-dependency (we didn't realise this when updating fn_go).

msgodf:
Looks good - to clarify, apparently this is the way to update a dep version: https://github.com/golang/dep/blob/master/docs/FAQ.md#how-do-i-change-the-version-of-a-dependency (we didn't realise this when updating fn_go).

Gopkg.toml
```diff
@@ -34,7 +34,7 @@
 
 [[constraint]]
   name = "github.com/fnproject/fn_go"
-  version = "0.2.7"
+  version = "0.2.8"
```
zootalures: You should run a dep-ensure locally to update the lock file and check that in  (otherwise dep ensure --vendor-only craps itself)

JadeRedworth: Instead of changing it manually?

PR #320, opened 2018-06-28
Added colours and fonts to the Help Text

vittoriaplacidi:
- Added bold style to the categories and the red colour to the title and link to follow the same colour pattern of the Fn logo
- Added italic style to some descriptions (further help, environment variables)
- Decided to have cyan colour as a colour pattern for each command


zootalures:
Colors are good but but this approach isn't portable  (e.g. to windows or other PTY types)  - typically color is only enabled in a CLI if the current stdout is a PTY (e.g. via https://github.com/mattn/go-isatty)  - color codes themselves aren't fully standandaridsed  also so this can be tricky .  

e.g. Typically 

./cmd_with_colors | cat 

would not produce tty color codes in its out 

but 

./cmd_with_colors 

would

michael-w-williams:
I like how it looks. Should look good on Dark Themes, which seem to be all the rage. The question is, will it create issues on some platforms. I will defer to @zootalures on that subject.

main.go
```diff
@@ -189,13 +191,27 @@ func commandArgOverrides(c *cli.Context) {
...
 func main() {
-	app := newFn()
 
-	err := app.Run(os.Args)
-	if err != nil {
-		// TODO: this doesn't seem to get called even when an error returns from a command, but maybe urfave is doing a non zero exit anyways? nope: https://github.com/urfave/cli/issues/610
-		fmt.Fprintf(os.Stderr, "ERROR: %v\n", err)
-		fmt.Fprintf(os.Stderr, "Client version: %s\n", Version)
-		os.Exit(1)
+	if tty.IsTerminal(os.Stdout.Fd()) || tty.IsCygwinTerminal(os.Stdout.Fd()) {
```
zootalures: I don't see any difference between these branche, what am i missing?

main.go
```diff
@@ -46,80 +47,81 @@ func newFn() *cli.App {
...
+		{{"\t"}}` + "\x1b[1mFURTHER HELP:\x1b[0m" + `{{"\t"}}` + "\x1b[3mSee \x1b[0m" + `'` + "\x1b[96;21mfn <command> --help\x1b[0m" + `' ` + "\x1b[3mfor more information about a command.\x1b[0m" + `{{if not .ArgsUsage}}
+	
+		{{"\t"}}` + "\x1b[1mLEARN MORE:\x1b[0m" + `{{"\t"}}{{"\t"}}` + "\x1b[91;4mhttps://github.com/fnproject/fn\x1b[0m" + `{{else}}{{end}}
 	`
-	//Override command template
+	// Override command template
 	// SubcommandHelpTemplate is the text template for the subcommand help topic.
 	// cli.go uses text/template to render templates. You can
 	// render custom help text by setting this variable.
 	cli.SubcommandHelpTemplate = `{{range .VisibleCategories}}{{if .Name}}
```
zootalures: My suggestion here is to put the coloring/highlighting in a function:

```
package color

var useColors  bool 
func init(){
   useColors = tty.IsTerminal(os.Stdout.Fd()) || tty.IsCygwinTerminal(os.Stdout.Fd()
}


func Bold(text string) string {
    if isColory {
         return "\x1b[1m" + text +  "\x1b[0m"
    }else { 
        return text 
   } 
}

to get this to work in a template you'll have to create a function map which points at those functions

**Comment by shaunsmith** [comment]
I was kind of hoping we could single source both help text output by the CLI and the online version by writing in Markdown and using a toolchain with [Pandoc](https://pandoc.org/) and a [Go static file embedding utility/framework](https://mattjibson.com/esc/).   Of course markdown doesn't support coloured fonts.

**Comment by shaunsmith** [comment]
Hm, maybe I've got this the wrong way around.  Maybe we single source from something richer than Markdown and use pandoc to generate Markdown for docs and something else for embedded help?

**Comment by vittoriaplacidi** [comment]
fn command help text

This is now how the 'fn' command help text looks, implemented following the @zootalures suggestion.

**Comment by vittoriaplacidi** [comment]
This is now how the 'fn create' & 'fn create app' commands help text looks, implemented following the @zootalures suggestion.

fn create command help text
fn create app command help text

**Comment by vittoriaplacidi** [comment]
Customised FURTHER HELP section for specific command name, instead of a general text.
screen shot 2018-07-12 at 11 48 15

File: main.go [hunk 3]
```diff
@@ -134,9 +80,31 @@ func newFn() *cli.App {
...
 	prepareCmdArgsValidation(app.Commands)
 
+	cli.HelpPrinter = func(w io.Writer, templ string, data interface{}) {
+		doStuff(w, templ, data, colour.Colours)
+	}
+
 	return app
 }
 
+func doStuff(out io.Writer, templ string, data interface{}, customFunc map[string]interface{}) {
```
zootalures: doStuff needs a better name

common/colour/colour.go
```diff
@@ -0,0 +1,100 @@
+package colour
```
zootalures: On the whole in code,  American spellings are prefered and we try and strictly adhere to this in Fn. - s/colour/color/

common/colour/colour.go
```diff
@@ -0,0 +1,100 @@
...
+import (
+	"os"
+	"strings"
+
+	tty "github.com/mattn/go-isatty"
+)
+
+var useColors bool
+
+var Colours map[string]interface{}
```
zootalures: s/Colours/Colors/

main.go
```diff
@@ -47,79 +51,21 @@ func newFn() *cli.App {
...
...talic .Usage}}{{"\n"}}{{"\n"}}{{"\t"}}{{bold "USAGE"}}{{"\n"}}{{"\t"}}{{"\t"}}{{boldcyan .HelpName}}{{if .VisibleFlags}}{{cyan " [global options]"}}{{end}} {{if .ArgsUsage}}{{brightred .ArgsUsage}}{{end}}{{if .Flags}}{{yellow " [command options]"}}{{end}}{{if .Description}}{{"\n"}}{{"\n"}}{{"\t"}}{{bold "DESCRIPTION"}}{{"\n"}}{{"\t"}}{{"\t"}}{{.Description}}{{end}}{{end}}{{end}}{{range .VisibleCategories}}{{if .Name}}{{"\n"}}{{"\n"}}{{"\t"}}{{bold .Name}}{{end}}{{range .VisibleCommands}}{{"\n"}}{{"\t"}}{{"\t"}}{{join .Names ", "}}{{"\t"}}{{"\t"}}{{"\t"}}{{.Usage}}{{end}}{{end}}{{if .VisibleFlags}}{{"\n"}}{{"\n"}}{{"\t"}}{{if not .ArgsUsage}}{{bold "GLOBAL OPTIONS"}}{{else}}{{bold "COMMAND OPTIONS"}}{{end}}{{"\n"}}{{"\t"}}{{"\t"}}{{range $index, $option := .VisibleFlags}}{{if $index}}{{"\n"}}{{"\t"}}{{"\t"}}{{end}}{{$option}}{{end}}{{end}}{{"\n"}}{{"\n"}}{{"\t"}}{{bold "FURTHER HELP:"}}{{"\t"}}{{italic "See "}}{{"'"}}{{brightcyan .HelpName}}{{brightcyan " <command> --help"}}{{"'"}}{{italic " for more information about a command."}}{{if not .ArgsUsage}}{{"\n"}}{{"\n"}}{{"\t"}}{{bold "LEARN MORE:"}}{{"\t"}}{{"\t"}}{{underlinebrightred "https://github.com/fnproject/fn"}}{{else}}{{end}}
```
zootalures: I don't think you need template commands like : `{{"\n"}}{{"\n"}}{{"\t"}}` in this template - normally you'd just include them as text in the raw template.

main.go
```diff
@@ -134,9 +80,31 @@ func newFn() *cli.App {
...
+	cli.HelpPrinter = func(w io.Writer, templ string, data interface{}) {
+		doStuff(w, templ, data, colour.Colours)
+	}
+
 	return app
 }
 
+func doStuff(out io.Writer, templ string, data interface{}, customFunc map[string]interface{}) {
+	funcMap := colour.Colours
+	if customFunc != nil {
```
zootalures: You don't need to test for nil on a golang map unless you are writing to it. 

when a map or array are  nil golang treats them as empty for iteration

main.go
```diff
@@ -134,9 +80,31 @@ func newFn() *cli.App {
...
 
+	cli.HelpPrinter = func(w io.Writer, templ string, data interface{}) {
+		doStuff(w, templ, data, colour.Colours)
+	}
+
 	return app
 }
 
+func doStuff(out io.Writer, templ string, data interface{}, customFunc map[string]interface{}) {
+	funcMap := colour.Colours
```
zootalures: do you need to be  passing the color map in as an arg and also using as the default values here?

common/colour/colour.go
```diff
@@ -0,0 +1,100 @@
+package colour
+
+import (
+	"os"
+	"strings"
+
+	tty "github.com/mattn/go-isatty"
```
zootalures: Generally you want to avoid using import aliases in golang  unless there is a conflict (the idea is that it's more farmiliar to readers when packages are always imported via the same prefix) 

Go ignores the `go-` prefix on the import so 
```
import  "github.com/mattn/go-isatty"
...
isatty.IsTerminal(...)
```

main.go
vittoriaplacidi: https://github.com/urfave/cli/blob/master/help.go#L236

I overwrite this function which is used to customise the AppTemplate.

PR #327, opened 2018-07-09
Sort subcommands alphabetically

JadeRedworth:


JadeRedworth:
The help command is provided by urfave, and now that we generate the subcommands ourselves this is the only way to order them. Else they are random and if you keep hitting `fn create --help` they will keep changing the order. So I'm not how we order the 'help' command.

PR #328, opened 2018-07-12
Fixed singular and plural for category names

vittoriaplacidi:
Fixed singular and plural for category names of the management commands.


vittoriaplacidi:
Here is an example of this small issue.
screen shot 2018-07-12 at 10 52 01
Singular category name is now shown for the help text of a command. @zootalures 
screen shot 2018-07-12 at 10 55 07

PR #335, opened 2018-07-17
Add leading '/' to trigger source on creation

JadeRedworth:


zootalures:
would be good to validate this on the server too.

PR #347, opened 2018-07-23
Add inspect command for context

JadeRedworth:
Inspect will print out all information held inside the provided context file.

`fn inspect context <context>`


JadeRedworth:
@msgodf yep makes sense, let me change!

msgodf:
One last point (honestly) - would it be worth telling the user what the current context is set to?

PR #348, opened 2018-07-23
Commands helptext

vittoriaplacidi:
Added description to management and server commands.
Fixed the usage description order of global options: added Trim function to color.go.


func.yaml
```diff
@@ -1,3 +1,8 @@
+schema_version: 20180708
```
JadeRedworth: You can delete these files, they are the result of `fn init` in your working directory.

func.yaml.bak
```diff
@@ -0,0 +1,3 @@
+name: cli
```
JadeRedworth: Same as above ^

objects/call/commands.go
```diff
@@ -12,7 +12,7 @@ func Get() cli.Command {
 		Usage:       "Get function call info per app",
 		Aliases:     []string{"calls", "cl"},
 		Category:    "MANAGEMENT COMMAND",
-		Description: "This is the description",
+		Description: "This command gets information call of an app.",
```
JadeRedworth: `This command gets information call of a route`

objects/context/commands.go
```diff
@@ -62,32 +64,35 @@ func Inspect() cli.Command {
...
-		Aliases:   []string{"ctx"},
-		ArgsUsage: "<key> <value>",
-		Category:  "MANAGEMENT COMMAND",
-		Action:    ctxMap.update,
+		Name:        "context",
+		Usage:       "Update context files",
+		Aliases:     []string{"ctx"},
+		ArgsUsage:   "<key> <value>",
+		Category:    "MANAGEMENT COMMAND",
+		Description: "This command updates context files.",
```
JadeRedworth: `This command updates the current context file.`

Issue #341, opened 2018-07-18
Print contextual info on failure


shaunsmith:
When the CLI fails to communicate with the server or with a repository or gets an error response code it would be useful to dump contextual info to remind the user of the current config, perhaps only when --verbose mode is enabled.  This would be helpful when users have changed contexts or repositories and commands are failing.  Useful contextual information could include:

- docker userid (who is the user logged in as?)
- value of FN_REGISTRY
- name of the current context
- any context refining provider specific values, like a profile name

Issue #330, opened 2018-07-12
As an application developer I would like the CLI to inform me when it has been updated and if I need to upgrade my server


rikgibson:
In short, when the CLI is upgraded, it would be good if the first time it is launched post-upgrade it informed the user that a new version is being used, display a link out to relevant Release Notes and output the recommended minimum server version to be used. Something like:

"Welcome to Fn CLI v0.4.8 - please check the release notes at http://github.com/fnproject/cli/README.md"

"Recommended minimum Fn Server version for use with this CLI is 0.3.603"

`fn version` could also display minimum server version warning if the version detected is too low.

rikgibson:
The CLI could cache it's version in `~/.fn/config.yaml` or a `~/.fn/.last-version.txt`and compare against it on launch, and if different then emit helpful messages before updating the cached version.

rikgibson:
Done - not doing anything about minimum server versions at this time, maybe later.

PR #351, opened 2018-07-23
Add contextual information for --verbose and first launch after upgrades.

JadeRedworth:
Related to issue: [#341](https://github.com/fnproject/cli/issues/341) Print contextual info on failure.

When `--verbose` is provided, print out the values of `FN_REGISTRY` and the current-context. 

Related to issue: [#330](https://github.com/fnproject/cli/issues/330) As an application developer I would like the CLI to inform me when it has been updated and if I need to upgrade my server.

Add a key/value `cli-version` into `~/.fn/config.yaml` to cache the version of the CLI the user has. When they update to a new version, the current version and the cached version in the config.yaml will be checked. If they differ the CLI will print out:
```
Welcome to Fn CLI 0.4.133 - please check the release notes at:
https://github.com/fnproject/cli/blob/master/README.md
```

Then upgrade the cached version in the config.yaml, so that this is only printed out on the first-time launch of a upgraded CLI.


commands/version.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/fnproject/cli/config"
+	"github.com/urfave/cli"
+)
+
+// VersionCommand
+func VersionCommand() cli.Command {
+	return cli.Command{
+		Name:        "version",
+		Usage:       "Display CLI and server versions",
+		Description: "This is commands shows the latest client and server version.",
```
rikgibson: "This command shows the version of the Fn CLI being used and the version of the Fn Server referenced by the current context, if available."

commands/version.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/fnproject/cli/client"
+	"github.com/fnproject/cli/config"
+	"github.com/urfave/cli"
+)
+
+// VersionCommand
+func VersionCommand() cli.Command {
+	return cli.Command{
+		Name:        "version",
+		Usage:       "Display CLI and server versions",
```
rikgibson: "Display Fn CLI and Fn Server versions"

common/common.go
```diff
@@ -109,6 +110,21 @@ func localBuild(path string, steps []string) error {
...
+	var registry, currentContext string
+	registry = viper.GetString(config.EnvFnRegistry)
+	if registry == "" {
+		registry = "fn registry is not set"
+	}
+	fmt.Println("FN_REGISTRY: ", registry)
+
+	currentContext = viper.GetString(config.CurrentContext)
+	if currentContext == "" {
+		currentContext = "no context currently in use"
```
rikgibson: No...

common/common.go
```diff
@@ -109,6 +110,21 @@ func localBuild(path string, steps []string) error {
 	return nil
 }
 
+func PrintContextualInfo() {
+	var registry, currentContext string
+	registry = viper.GetString(config.EnvFnRegistry)
+	if registry == "" {
+		registry = "fn registry is not set"
```
rikgibson: FN_REGISTRY is not set

config/config.go
```diff
@@ -128,9 +130,16 @@ func LoadConfiguration(c *cli.Context) error {
 	viper.AddConfigPath(filepath.Join(home, rootConfigPathName, contextsPathName))
 	viper.SetConfigName(context)
 
+	if viper.GetString("cli-version") != "" {
```
rikgibson: Why only output this "Welcome..." message if the version isn't set in config? I'd output it if the version is missing as well as if it's new - both indicate a new version essentially.

config/config.go
```diff
@@ -128,9 +130,16 @@ func LoadConfiguration(c *cli.Context) error {
...
+		if viper.GetString("cli-version") != Version {
+			fmt.Printf("Welcome to Fn CLI %s - please check the release notes at:\n%s \n\n", Version, "https://github.com/fnproject/cli/blob/master/README.md")
+		}
+	}
+	WriteCurrentContextToConfigFile(CurrentCliVersion, Version)
```
rikgibson: This doesn't seem to actually write anything - my config.yaml is unchanged.

Also, I'd change the name of "WriteCurrentContextToConfigFile" if the function is writing more than just the current context - something like "WriteConfigValueToConfigFile" maybe?

config/config.go
```diff
@@ -142,7 +151,7 @@ func LoadConfiguration(c *cli.Context) error {
 	return nil
 }
 
-func WriteCurrentContextToConfigFile(value string) error {
+func WriteCurrentContextToConfigFile(key, value string) error {
```
rikgibson: Ah yes, this function should have a better name now.

Issue #343, opened 2018-07-19
fn build usage is inconsistent with other command as it only works in the function directory


delabassee:
Main Fn commands (`test`, `run`, `deploy`) can be invoked using a path to the function directory or directly in the function directory while `fn build` only works in the function directory.

`fn init --runtime go myFunc`
`fn build myFunc` -> NOK
`fn test myfunc` -> OK
`fn run myFunc` -> OK
`fn deploy --app test --local myFunc` -> OK

PR #352, opened 2018-07-24
Add path option to build command

JadeRedworth:
[#343](https://github.com/fnproject/cli/issues/343) fn build usage is inconsistent with other command as it only works in the function directory

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -2,6 +2,8 @@ package commands
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 
 	"github.com/fnproject/cli/common"
 	"github.com/urfave/cli"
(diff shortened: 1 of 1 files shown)


msgodf:
Does this clash/duplicate this argument?
`--working-dir value, -w value  Specify the working directory to build a function, must be the full path.`

msgodf:
The help text for `--build` doesn't tell me about this way of using it.

JadeRedworth:
It's the same behavior that `run` `test` `deploy` and `bump` have. The `--working-dir` flag specifies the chosen working directory, mainly required for CI. The path then gets appended to the current working directory. Whether than provided by the `--working-dir` or retrieved in the code.

msgodf:
Okay, so the help text for all of them isn't helpful enough then :) this is consistent. I'll create an issue to make the help text say that `fn COMMAND [global options]  [command options]` accepts a directory name.

JadeRedworth:
Ahh I see, you mean for the "path". Yes, things need to be changed as it's not clear that a user can even put this arg.

Issue #344, opened 2018-07-19
`fn call` doesn't log the actual call with `DEBUG=1`


riconnon:
I would expect to see output for the HTTP headers, etc the same as I do for API calls

rikgibson:
Fixed and closed by @mantree

PR #355, opened 2018-07-24
Adding debug logging for call request/response #344

mantree:
The Call api endpoint isn't included in the Swagger
definition, so we hand build a http client. This doesn't include the
debug creation as that is performed by the Swagger Runtime code.

This change copies the logging approach from the Runtime code, adding
a context check for 'verbose' mode as well.

### DIFF
--- a/client/call_fn.go
+++ b/client/call_fn.go
@@ -7,10 +7,12 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/http/httputil"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


mantree:
@rikgibson This change is to make `fn call` work like all the other invocations already do. I'm not going to merge it with the 'verbose' flag, that just to test it really.

@JadeRedworth have you got some commits that show what your up to in this area? Is 'verbose' going to enable logger.DebugEnabled()?

PR #357, opened 2018-07-26
Save space in python FN runtime image

shurkaxaa:
Save space in python runtime env by removing packages required for pip install after installation,  apt cache cleanup and pip install related layers merge into single one.
As result python sample image reduced from 360Mb to 165Mb
from which 135Mb is base python env image.

### DIFF
--- a/langs/python.go
+++ b/langs/python.go
@@ -71,17 +71,17 @@ func (h *PythonLangHelper) Entrypoint() (string, error) {
 }
 
 func (h *PythonLangHelper) DockerfileBuildCmds() []string {
-	pip := "pip3"
 	r := []string{}
-	r = append(r, "RUN apt-get update && apt-get install --no-install-recommends -qy build-essential gcc")
-	if exists("requirements.txt") {
-		r = append(r,
-			"ADD requirements.txt /function/",
-			fmt.Sprintf("RUN %v install --no-cache --no-cache-dir -r requirements.txt", pip),
-		)
-	}
 	r = append(r, "ADD . /function/")
-	r = append(r, "RUN rm -fr ~/.cache/pip /tmp* requirements.txt func.yaml Dockerfile .venv")
+	if exists("requirements.txt") {
+		r = append(r, `
+RUN apt-get update && apt-get install --no-install-recommends -qy build-essential gcc &&\
+    pip3 install --no-cache --no-cache-dir -r requirements.txt &&\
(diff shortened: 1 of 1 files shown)


delabassee:
@shurkaxaa 's OCA has been reviewed and approved

langs/python.go
```diff
@@ -71,17 +71,17 @@ func (h *PythonLangHelper) Entrypoint() (string, error) {
...
 	r := []string{}
-	r = append(r, "RUN apt-get update && apt-get install --no-install-recommends -qy build-essential gcc")
+	r = append(r, "ADD . /function/")
 	if exists("requirements.txt") {
-		r = append(r,
-			"ADD requirements.txt /function/",
-			fmt.Sprintf("RUN %v install --no-cache --no-cache-dir -r requirements.txt", pip),
-		)
+		r = append(r, `
+RUN apt-get update && apt-get install --no-install-recommends -qy build-essential gcc &&\
```
zootalures: I think this change is fine now but I think the python FDK build itself is wrong to be running  apt commands by default during fn builds - I've raised this to track: https://github.com/fnproject/fdk-python/issues/44

zootalures:
Not sure why the CLA bot isn't picking this up, verified as signed.

langs/python.go
denismakogon: The way this change was introduced makes the whole build process slower than it was before. With this change package install process will be triggered with any change to *.py files, but it’s wrong. The only reason to trigger package install process is change to requirements.txt

So, this change has to be rewritten to reduce build time as it was before, but with new purge thing.

shurkaxaa: I believe proper way will be just to create apt repo and have pre-built package like python3-fdk for stretch. In that case no compile or other installed packages will be required during build. Or in ideal case base image shall have FDK installed so build process will just put func.py to image and install custom requirements from requirements.txt if any. And even in that case you need to have build essentials only for native pip requirements, pure python packages does not need it.

denismakogon: I tend to disagree with you because build packages can be required by other dependencies that require compiling (tensorflow, numpy, etc.)

shurkaxaa: They also can require devel packages too (like openssl-devel), so keeping just compiler will not help anyway

shurkaxaa: I think best solution from size point of view - multistage images, so release image will not have unnecessary things. But image creation time will be much longer

PR #358, opened 2018-07-26
Python multistage support

denismakogon:
dev
===
New managed base image is `fnproject/python:3.6-dev`.
OS: `Debian`
Base image: `python:3.6-slim-stretch`
Installed packages: `build-essential gcc`

New managed base image is `fnproject/python:3.7-dev`.
OS: `Debian`
Base image: `python:3.7-slim-stretch`
Installed packages: `build-essential gcc`

runtime
======
New managed base image is `fnproject/python:3.6`.
New managed base image is `fnproject/python:3.7`.

### DIFF
--- a/langs/python.go
+++ b/langs/python.go
@@ -59,11 +59,11 @@ func (h *PythonLangHelper) Extensions() []string {
 }
 
 func (h *PythonLangHelper) BuildFromImage() (string, error) {
-	return fmt.Sprintf("python:%s-slim-stretch", h.Version), nil
+	return fmt.Sprintf("fnproject/python:%s", h.Version), nil
 }
 
 func (h *PythonLangHelper) RunFromImage() (string, error) {
-	return fmt.Sprintf("python:%s-slim-stretch", h.Version), nil
+	return fmt.Sprintf("fnproject/python:%s", h.Version), nil
 }
 
 func (h *PythonLangHelper) Entrypoint() (string, error) {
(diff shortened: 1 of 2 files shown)


denismakogon:
Note: new images is not on dockerhub yet.

langs/python.go
```diff
@@ -75,8 +75,7 @@ func (h *PythonLangHelper) DockerfileBuildCmds() []string {
 	r = append(r, "ADD . /function/")
 	if exists("requirements.txt") {
 		r = append(r, `
-RUN apt-get update && apt-get install --no-install-recommends -qy build-essential gcc &&\
-    pip3 install --no-cache --no-cache-dir -r requirements.txt &&\
+RUN pip3 install --target /python/  --no-cache --no-cache-dir -r requirements.txt &&\
     apt-get remove -y --purge build-essential gcc && apt-get autoremove -y &&\
```
zootalures: Think you can drop the following three lines as this doesn't install packages any more.

langs/python.go
```diff
@@ -75,8 +75,7 @@ func (h *PythonLangHelper) DockerfileBuildCmds() []string {
 	r = append(r, "ADD . /function/")
 	if exists("requirements.txt") {
 		r = append(r, `
-RUN apt-get update && apt-get install --no-install-recommends -qy build-essential gcc &&\
-    pip3 install --no-cache --no-cache-dir -r requirements.txt &&\
+RUN pip3 install --target /python/  --no-cache --no-cache-dir -r requirements.txt &&\
```
zootalures: I think you should add `--disable-pip-version-check` here  as it drops the spam and speeds up pip a tiny bit

langs/python.go
```diff
@@ -75,8 +75,7 @@ func (h *PythonLangHelper) DockerfileBuildCmds() []string {
 	r = append(r, "ADD . /function/")
 	if exists("requirements.txt") {
```
zootalures: not 100% sure it's right but I think you can save running pip every time by pre-loading requirements.txt and running pip before copying the rest of the functions context 

```
r := []string{}
	if exists("requirements.txt") {
		r = append(r, "ADD requirements.txt /function/requirements.txt")
		r = append(r, `
RUN pip3 install --target /python/  --disable-pip-version-check  --no-cache --no-cache-dir -r requirements.txt &&\
    rm -fr ~/.cache/pip /tmp* requirements.txt func.yaml Dockerfile .venv`)
	}

	r = append(r, "ADD . /function/")

	return r
```

zootalures: on my mac at least this lets the pip stuff get cached in docker build.

PR #359, opened 2018-07-26
List triggers

vittoriaplacidi:
Added support to list triggers by app


objects/trigger/triggers.go
```diff
@@ -112,25 +114,36 @@ func CreateTrigger(client *clientv2.Fn, trigger *models.Trigger) error {
 func (t *triggersCmd) list(c *cli.Context) error {
 	appName := c.Args().Get(0)
 	fnName := c.Args().Get(1)
+	var params *apitriggers.ListTriggersParams
```
rikgibson: small code format thing - `apitriggers` should be camel-cased to `apiTriggers`

objects/trigger/triggers.go
```diff
@@ -112,25 +114,36 @@ func CreateTrigger(client *clientv2.Fn, trigger *models.Trigger) error {
...
+		}
+		params = &apitriggers.ListTriggersParams{
+			Context: context.Background(),
+			AppID:   &app.ID,
+			FnID:    &fn.ID,
+		}
 	}
 
 	var resTriggers []*models.Trigger
+	//var resFns []*models.Fn
```
rikgibson: Remove this if resFns isn't needed.

Issue #356, opened 2018-07-25
Specify templates for `fn init --runtime=.....` outside of the CLI source code


mjg123:
As a function developer I want to specify my own template when creating a new function so that I can provide reduce the amount of boilerplate updating.
As an FDK developer I want to not have my FDK coupled to the CLI by the hard-coded project templates in the CLI.

Strawman suggestion for how this could work:

  - Have each supported runtime use an "init-image" that the CLI pulls and invokes when generating a new function.
  - User can specify a different "--init-image=<image>" to plug in their own boilerplate.
  - The contract for images is to run with some arguments such as `function-name` (and which others?) and to stream out a tarball of the function template, which the CLI can put in place on disk. The alternative of mounting the directory in is risky because of the difficulty in reliably determining the correct user.

shaunsmith:
I have previously suggested a configurable Docker image based system with a defined contract between CLI and image so any number of files could be return as part of the template.  @rikgibson did we capture this as issues or just in meeting notes?

rikgibson:
Hey @shaunsmith I captured the overall concept as a JIRA, which is where this issue ultimately spawned from. We didn't really get into the nitty gritty though - what's the contract, online vs offline experience, breadth of language support etc.  I'd also like to see the build and run images built by the FDK (I think the Java FDK images already are) and the values of build_image and run_image in func.yaml populated by the FDK based on params supplied to `fn init`, rather than baking into CLI.

rikgibson:
Possible params to configure the boilerplate, other than `function-name` - `class-name`, `language-version`? Others? `language-version` may or may not entail different boilerplate, certainly different build/run images.

mjg123:
I'll prototype this and see what I find.

shaunsmith:
We should also think beyond language and support arbitrary init/build/run images so that a team could define a template that matches their project needs or corporate standards, for example.  I had envisioned the CLI supporting registration of "runtimes" which is maintained in the `.fn/config.yaml` with defaults for supported runtimes.

PR #360, opened 2018-07-27
Added a support for GraalVM native-image.

tzezula:
Added a support for the GraalVM `native-image` into the FnProject `cli`.

The `SVMLangHelper` creates a FnProject Java Maven Project which is extended by a maven profile doing the native image build. It uses its own `Dockerimage`s for building and running.
The buid `Dockerimage` is based on `openjdk:8-jdk-slim` which is extended by open source GraalVM `native-image.
The run `Dockerimage` is based on `scratch`, the run images are small and fast to start.

This Pull Request depends on Docker Images from the Pull Request: https://github.com/fnproject/dockers/pull/15/
Question: Should I publish these images to Docker Hub or will they be created by infrastructure when the Pull Request is integrated?

Comments are welcomed.

### DIFF
--- a/langs/base.go
+++ b/langs/base.go
@@ -21,7 +21,7 @@ func init() {
 	registerHelper(&RubyLangHelper{})
 	registerHelper(&RustLangHelper{})
 	registerHelper(&KotlinLangHelper{})
-
+	registerHelper(&SVMLangHelper{javaHelper:JavaLangHelper{version: defaultSVMJavaSupportedVersion}})
 }
 
 func registerHelper(h LangHelper) {
--- a/langs/svm.go
+++ b/langs/svm.go
@@ -1,0 +1,248 @@
+package langs
+
+import (
+	"os"
+	"path/filepath"
+	"io/ioutil"
+	"strings"
+	"bufio"
+	"unicode"
+)
+
+type SVMLangHelper struct {
+	BaseHelper
+	javaHelper JavaLangHelper
+}
+
+func (svm *SVMLangHelper) Handles(language string) bool {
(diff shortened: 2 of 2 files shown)


rikgibson:
I don't think this is the right way to go about this. A couple of points:

1) The Java FDK now takes responsibility for producing the build and run images for JVM-based languages, so consider moving your image build out of fnproject/dockers into fnproject/fdk-java

2) I don't think SVM is really a valid runtime type - it's a VM choice that should be open to people building using any JVM-based language and probably should simply be a value option for the `run_image` in the func.yaml.

3) More strategically, these options should be folded into the proposal detailed in #356 - @mjg123 - such that variations in boilerplate for SVM should be dealt with there.

delabassee:
Unless I'm wrong SVM is still based on SE8 so we probably want to keep Java and SVM separate.

tzezula:
@rikgibson :

1. OK, I will move the Dockerfiles

2. Yes, I agree that the `native-image` can be used by any JVM based language.  But the `run_image` currently cannot be used. The problem is that jar files in `/function/app/` are not enough for the `native-image`. It also requires reflection configuration file which is user edited.
I am working on removing the need for it for FnProject entry point and Jackson reflection but it
maybe still needed for user code or external libraries.

3. Agree, making the `CLI` pluggable not to require the changes in CLI would be great.
Any estimate for the #356 PR.

rikgibson:
This should update to be in line with 
https://github.com/fnproject/cli/pull/369 when it's eventually done.

tzezula:
I've rewritten the feature on top of `fn init --init-image`.
This PR is no more needed.

PR #363, opened 2018-07-27
update go client version

rdallman:
I've had to run `dep ensure` upon updating this repo and it consequently
updated my `fn_go` client version and the `fdk-go` dependency as well. Figured
I would check these in instead of having to stash/update them across branches.


rdallman:
thanks!

PR #368, opened 2018-08-05
Clojure FDK

jgh-:
Hello,  I have created a Clojure FDK from code I've been working on for my project.  The FDK itself can be found at https://github.com/unpause-live/fdk-clj.  It currently supports CloudEvents and JSON format and passes Fn Test.

In this PR is a Clojure language helper which will, as with the other languages, prepare some example boilerplate.   The supplied tests in the boilerplate pass as well.

In the current example I have it building with `clojure:lein` and running with `openjdk:9-jre`, however functions could also be built using Graal's `native-image` and run from `SCRATCH`.

### DIFF
--- a/langs/base.go
+++ b/langs/base.go
@@ -21,6 +21,7 @@ func init() {
 	registerHelper(&RubyLangHelper{})
 	registerHelper(&RustLangHelper{})
 	registerHelper(&KotlinLangHelper{})
+	registerHelper(&ClojureLangHelper{version: "1.9.0"})
 
 }
 
--- a/langs/clojure.go
+++ b/langs/clojure.go
@@ -1,0 +1,249 @@
+package langs
+
+import (
+    "bytes"
+    "encoding/json"
+    "errors"
+    "fmt"
+    "io/ioutil"
+    "net/http"
+    "os"
+    "path/filepath"
+)
+
+// ClojureLangHelper provides a set of helper methods for the lifecycle of Clojure Maven projects
+type ClojureLangHelper struct {
+    BaseHelper
+    latestFdkVersion string
+    version          string
+}
+
+func (h *ClojureLangHelper) Handles(lang string) bool {
+    for _, s := range h.LangStrings() {
+        if lang == s {
+            return true
+        }
+    }
+    return false
+}
+func (h *ClojureLangHelper) Runtime() string {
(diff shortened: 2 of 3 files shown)


langs/clojure.go
```diff
@@ -0,0 +1,249 @@
...
+        fullPath := filepath.Join(path, dir)
+        if err = os.MkdirAll(fullPath, os.FileMode(0755)); err != nil {
+            return err
+        }
+
+        fullFilePath := filepath.Join(fullPath, filename)
+        return ioutil.WriteFile(fullFilePath, []byte(content), os.FileMode(0644))
+    }
+
+    err = mkDirAndWriteFile("src/example", "core.clj", helloClojureSrcBoilerplate)
```
denismakogon: let's call it `src/func`, we `func` across multiple lang helpers, so, let's stick with that

langs/base.go
```diff
@@ -21,6 +21,7 @@ func init() {
 	registerHelper(&RubyLangHelper{})
 	registerHelper(&RustLangHelper{})
 	registerHelper(&KotlinLangHelper{})
+	registerHelper(&ClojureLangHelper{version: "1.9.0"})
```
denismakogon: overall, looks good, need to do few manual tests.

in a meantime, can you please add tests for this helper?
take a look at:
 - https://github.com/fnproject/cli/blob/master/test/cli_lang_boilerplate_test.go#L15

jgh-: Sure, will do.

denismakogon:
Would be nice to write an article on medium about that.

jgh-:
OK, I'm not much of a blogger but happy to write an article about it!  It's something I should be doing more of anyway.

mjg123:
Yes - agree.  I'm aiming to get the init-image stuff at #369 merged asap so that this can be supported.  I do ❤️ clojure after all 😄

jgh-:
OK, I will revisit this when #369 is merged and figure out the correct way for the Clojure FDK to be compliant!

mjg123:
We're documenting it now, but roughly what we will need is a docker image which will be run with something like:  `docker run -e FN_FUNCTION_NAME=<the_name> <your_image>`.  The output should be something which can be piped to `tar -x` to produce the files on disk.  We will be creating several examples which you can copy from.

PR #369, opened 2018-08-08
init-images

vittoriaplacidi:
This command:

```
  fn init --image=mjg123/java-init:latest
```

Will `docker run` that image, un-tarring the output with `archive/tar`.

The expected result is that a template of the function will be created, along
with a `func.init.yaml` which is missing various field (notably `Name` and `Version`).  The missing fields are then added by the cli, so that the function is ready to `fn run`.

### DIFF
--- a/commands/init.go
+++ b/commands/init.go
@@ -223,7 +223,7 @@ func (a *initFnCmd) init(c *cli.Context) error {
 
 		// Merge the func.yaml from the initImage with a.ff
 		//     write out the new func file
-		var initFf, err = common.ParseFuncfile("func.yaml")
+		var initFf, err = common.ParseFuncfile("func.init.yaml")
 		if err != nil {
 			return errors.New("init-image did not produce a valid func.yaml fragment")
 		}
--- a/commands/init.go
+++ b/commands/init.go
@@ -20,9 +20,11 @@ it will print an error message then exit.
 */
 
 import (
+	"archive/tar"
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -197,35 +199,29 @@ func (a *initFnCmd) init(c *cli.Context) error {
 		fmt.Println("Building from init-image: " + initImage)
 
 		// Run the initImage
-		var c1ErrB, c2ErrB bytes.Buffer
+		var c1ErrB bytes.Buffer
+		tarR, tarW := io.Pipe()
 
 		c1 := exec.Command("docker", "run", "-e", "FN_FUNCTION_NAME="+a.ff.Name, initImage)
 		c1.Stderr = &c1ErrB
+		c1.Stdout = tarW
 
-		c2 := exec.Command("tar", "-x")
-		c2.Stderr = &c2ErrB
-		c2.Stdin, _ = c1.StdoutPipe()
-		c2.Stdout = os.Stdout
-
-		_ = c2.Start()
-		c1_err := c1.Run()
-		c2_err := c2.Wait()
-
+		c1_err := c1.Start()
 		if c1_err != nil {
 			fmt.Println(c1ErrB.String())
 			return errors.New("Error running init-image")
 		}
 
-		if c2_err != nil {
-			fmt.Println(c2ErrB.String())
-			return errors.New("Error un-tarring output from init-image")
+		err = untarStream(tarR)
+		if err != nil {
+			return errors.New("Error un-tarring the output of the init-image")
 		}
 
 		// Merge the func.yaml from the initImage with a.ff
 		//     write out the new func file
 		var initFf, err = common.ParseFuncfile("func.init.yaml")
 		if err != nil {
-			return errors.New("init-image did not produce a valid func.yaml fragment")
+			return errors.New("init-image did not produce a valid func.init.yaml")
 		}
 
 		initFf.Name = a.ff.Name
@@ -234,6 +230,8 @@ func (a *initFnCmd) init(c *cli.Context) error {
 		if err := common.EncodeFuncfileYAML("func.yaml", initFf); err != nil {
 			return err
 		}
+
+		os.Remove("func.init.yaml")
 
 	} else {
 
@@ -252,10 +250,50 @@ func (a *initFnCmd) init(c *cli.Context) error {
 
 	}
 
-
-
 	fmt.Println("func.yaml created.")
 	return nil
+}
+
+// Untars an io.Reader into the cwd
+func untarStream( r io.Reader ) error {
+
+	tr := tar.NewReader(r)
+	for {
+		header, err := tr.Next()
+
+		if err == io.EOF {
+			// if no more files are found we are finished
+			return nil
+		}
+
+		if (err != nil){
+			return err
+		}
+
+		switch header.Typeflag {
+			// if its a dir and it doesn't exist create it
+			case tar.TypeDir:
+				if _, err := os.Stat(header.Name); err != nil {
+					if err := os.MkdirAll(header.Name, 0755); err != nil {
+						return err
+					}
+				}
+
+			// if it's a file create it
+			case tar.TypeReg:
+				f, err := os.OpenFile(header.Name, os.O_CREATE|os.O_RDWR, os.FileMode(header.Mode))
+				if err != nil {
+					return err
+				}
+
+				// copy over contents
+				if _, err := io.Copy(f, tr); err != nil {
+					return err
+				}
+
+			f.Close()
+		}
+	}
 }
 
 func (a *initFnCmd) initV2(c *cli.Context, fn modelsV2.Fn) error {
--- a/commands/init.go
+++ b/commands/init.go
@@ -61,7 +61,7 @@ func initFlags(a *initFnCmd) []cli.Flag {
 			Usage: "Choose an existing runtime - " + langsList(),
 		},
 		cli.StringFlag{
-			Name:  "image",
+			Name:  "init-image",
 			Usage: "A Docker image which will create a function template",
 		},
 		cli.StringFlag{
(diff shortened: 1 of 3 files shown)


mjg123:
BTW - in case it wasn't clear - DO NOT MERGE THIS as-is. It's a PoC and needs tests & docs at least.

The init image at mjg123/java-init was created from the hard-coded template in the CLI.  Dockerfile is:

```
FROM ubuntu:latest
RUN ["tar", "cvf", "java-init.tar", "*"]
CMD ["cat", "java-init.tar"]
```

(the ubuntu bit is obviously bigger than necessary.)

mjg123:
#356 is the issue for this

mjg123:
Params:  The function name and version can be passed into the init-image (as environment variables perhaps?). It's a nice touch to have the fn name in the code, but AFAIK none of the runtime templates do any customisation like that right now so I don't think it's a big deal to start with.

Language Version: As I see it, Java8/11/SVM are different runtimes and can have separate init-images. We should aim to keep the init-images as simple as possible - I wouldn't want to encourage much (any) logic in them.

delabassee:
If we don't support param/env to set the function name, default values (classes name, artifactId for Java) should be more sensible than they're today.

rikgibson:
Customising the classname was part of the original ask for this, e.g. `fn init --runtime java foo` wouldn't just create a `/foo` directory for the function, it would also create a class called `FooFunction` instead of `HelloFunction`.

Regarding different language versions, it depends on where you want the variant logic - in the FDK or in the CLI. SCM is maybe a special case, but more generally if [Java] 8 vs 9 vs 10 vs 11 vs 12 etc means  building a separate init image per version in the FDK and then adding more `if version = 8 else if version = 9 else if ...` cases into the CLI, vs a single more flexible init image and a simpler CLI.

Ultimately if we can support passing of env vars into the init image then each image is free to be as simple or as flexible as it wants.

shaunsmith:
This looks awesome!  But we need to consider the `build` step after `init` as the code layout created by init has to be buildable by the build image.  For example, if the Java init image generates a Gradle build it will have to be paired with an appropriate build image.  @mjg123 are you thinking that an init image author would use the `func.yaml` `build_image` property to specify the appropriate build image?  

And I agree with @rikgibson that we need a mechanism for passing customization parameters directly on the command line.  It's fine to have a fixed set of defined parameters with fixed names like FN_FUNCTION_NAME but it would be nice to offer authors of init images the ability to pass whatever they need.  That said, we don't need this support immediately and we could evaluate whether this is an actual need based on user feedback.   Function name is definitely something that has been asked for and should be included in the first release.

mjg123:
Yeah one problem here is that func.yaml is defining two separate things: build details and deployment details - and for some items it's not clear where the responsibility lies.

FWIW `fn init --runtime=java` and `fn init --init-image=mjg123/java-init` produce _exactly_ the same content with this patch (neither produces a `schema-version` attribute).  I've just lifted everything out of the hard-coded LangHelper into an external image.  It seems like we can add `init-image` as a replacement for the current functionality as-is without having to block it while we solve all the other problems we've just now noticed.

mjg123:
> I see the images as returning a func.yaml.contrib almost that the CLI will merge into whatever it wants to generate/include. I think the values of build_image and run_image should be part of this contribution as well - it shouldn't be up to the CLI to add those in.

this is exactly how it works right now, except the file is called `func.yaml` rather than `func.yaml.contrib`.  I'll change that right away because it is confusing having 2 files called the same thing.

rikgibson:
Hmmm, the lack of a `schema_version` is a bug - seems it only gets injected into the file if "--trigger http" is specified, e.g. `fn init --runtime java --trigger http foo-func`

mjg123:
So, here is a list of all the possible keys in `func.yaml`. Can we split them into "Provided by CLI" and "Provided by init-image" please?  I've taken a stab at it but happy to swap things around if I'm wrong:

Init-image:
  - build_image, cmd, content_type, entrypoint, expects, format, headers, run_image, runtime

CLI:
  - cpus, idle_timeout, memory, name, path, schema_version, timeout, triggers, version

Not sure:
  - build, config, schema_version, tests, type

rikgibson:
`fn init --help` spits out this:

```[~/test]: fn init --help

DEVELOPMENT COMMANDS
  fn init -   Create a local func.yaml file

USAGE
  fn [global options] init [command options] [function-subdirectory]

DESCRIPTION
  This command creates a func.yaml file in the current directory.

COMMAND OPTIONS
  --name value                   Name of the function. Defaults to directory name in lowercase.
  --force                        Overwrite existing func.yaml
  --runtime value                Choose an existing runtime - dotnet, go, java8, java9, java, lambda-nodejs4.3, lambda-node-4, node, php, python, python3.6, python, python3.7, ruby, rust, kotlin
  --entrypoint value             Entrypoint is the command to run to start this function - equivalent to Dockerfile ENTRYPOINT.
  --cmd value                    Command to run to start this function - equivalent to Dockerfile CMD.
  --version value                Set initial function version (default: "0.0.1")
  --working-dir value, -w value  Specify the working directory to initialise a function, must be the full path.
  --trigger value                Specify the trigger type.
  --memory value, -m value       Memory in MiB (default: 0)
  --type value, -t value         Route type - sync or async
  --config value, -c value       Route configuration
  --headers value                Route response headers
  --format value, -f value       Hot container IO format - default or http
  --timeout value                Route timeout (eg. 30) (default: 0)
  --idle-timeout value           Route idle timeout (eg. 30) (default: 0)
  --annotation value             Route annotation (can be specified multiple times)```

so at least some of the keys you've grouped into init-image can be overridden by the CLI, which I guess is fine - the image provides the default, and the user/CLI can override when merging in the contrib file.

File: commands/init.go [hunk 1]
```diff
@@ -180,18 +191,69 @@ func (a *initFnCmd) init(c *cli.Context) error {
...
+		c1.Stderr = &c1ErrB
+
+		c2 := exec.Command("tar", "-x")
+		c2.Stderr = &c2ErrB
+		c2.Stdin, _ = c1.StdoutPipe()
+		c2.Stdout = os.Stdout
+
+		_ = c2.Start()
+		c1_err := c1.Run()
+		c2_err := c2.Wait()
```
rdallman: normally I just use https://golang.org/pkg/os/exec/#Cmd.CombinedOutput cuz I'm super lazy, may clean up error handling/etc. here if it's possible to use it (didn't investigate further sorry)

rdallman:
This is great, thanks for doing it. 

my only concern is loose around `tar` usage. I expect it works the same across our primarily supported platforms -- reasonably up to date `win|osx|linux` -- but I'm maybe 90% I really don't know for sure (win, in particular, may 7zip reign supreme).

does this produce images [loaded into docker] exactly the same as previously that are named with the tagged version? it's not clear what artifact is produced from the pr alone, I haven't run this code -- pardon my stupidity. it's clear we can `fn run` them, my question is do we get a tar ball out of this or do we get an image in docker I could see in e.g. `docker images` (I don't see `docker load`)? are tar balls left laying around? if tar balls are left laying around, I'd prefer to see them cleaned up, if not for filesystem space concerns/bloat, this is some additional 'magic' we should hide from users imo, and many users may not know what these files are (many users not perfectly familiar with docker magic to begin with, this is another onion layer lower).

mjg123:
The init image is required to produce output on stdout in posix tar format. We pipe it to `tar -x` which works on win10/linux/mac.  Basically we do `docker run init-image | tar -x` to produce files on disk, just like other forms of `fn init` use the LangHelpers to do that. Nothing needs cleaning up afterwards.

commands/init.go
mjg123: For c1 (ie `docker run`) we need to handle them separately - stdout goes to c2.stdIn & stderr goes to the user.

For c2 (ie `tar -x`) we expect stdout to be empty.  Any errors should go to the user.

zootalures:
Would rather this used `archive/tar` than relying on an external tool

zootalures:
Just a thought - but have we considered ditching any kind of build magic completely and just generating a Dockerfile for all languages?  - this de couples the Fn cli from build (can still run docker build but absolutely no magic is involved)

mjg123:
> have we considered ditching any kind of build magic completely
 ---> #383

mjg123:
> Would rather this used archive/tar than relying on an external tool
done

JadeRedworth:
> Hmmm, the lack of a schema_version is a bug - seems it only gets injected into the file if "--trigger http" is specified, e.g. fn init --runtime java --trigger http foo-func

This was intended, we are using `schema_version` to detect if func.yaml is V1 or V2, rather than identifying an actual version.

shaunsmith:
With #383 proposing to eliminate `fn build` and to use `docker build` as part of a move to be more openly Docker-centric (which I'm cool with) I think we do actually need a general way to pass parameters to `fn init` to support logic that customizes the generated Dockerfile.   I was looking for a use case and then remembered my own. 😉 I've been updating the Java language helper to allow me to customize the generate Dockerfile to optionally use a Maven settings.xml and settings-security.xml.  With the Dockerfile generation moving into the init image I need a way to toggle this support.

mjg123:
> optionally use a Maven settings.xml and settings-security.xml. With the Dockerfile generation moving into the init image I need a way to toggle this support.

How do you imagine this looking for a user?  It's one thing to add a K/V property and a different thing to pass a file in from the host.

Also: How does this work at the moment?  Do we lose anything by moving to init-images or can this be added afterwards?

shaunsmith:
Of course the other solution is to not support "intelligent" Dockerfile generation and leave it to the user to edit the generated file themselves.  This isn't directly possible in the current CLI based solution because the Dockerfile isn't exposed.  But if we do expose it then customization is perfectly fine--which is kind of the point of your proposal!! 😃

mjg123:
Yes I definitely prefer putting less "intelligence" in the CLI - and then the init-images can be simpler too. Doockerfiles are way too complex for us to support parameterizing them in every way possible.  This is also discussed in #383

mjg123:
Did a rebase just now.  This is feature-complete but needs a better suite of tests and then docs about init-images will need updating.

testharness/harness.go
```diff
@@ -436,6 +436,16 @@ func (h *CLIHarness) MkDir(dir string) {
 
 }
 
+func (h *CLIHarness) Exec(name string, args ...string) error {
+	cmd := exec.Command(name, args...)
+	cmd.Dir = h.cwd
+	err := cmd.Run()
+	//	out, err := cmd.CombinedOutput()
```
JadeRedworth: can these be removed?

PR #379, opened 2018-08-10
Add callfn command which directly invokes a fn via the new invoke endpoint

mantree:
Changes:
 * Current call is left in place as a command, but internally it is renamed to 'CallRoute', to help make things clear between the two.
 *  Add callfn command which calls the new /invoke/fnID endpoint on the server, which is being added in fnproject/fn#1168


mantree:
Need to make this use fn name on the cli, not the id

rikgibson:
yeah, this should be `fn invoke <app-name> <func-name>`

riconnon:
I think we should have at least a basic test of this new feature.

commands/test.go
```diff
@@ -277,7 +277,7 @@ func (t *testcmd) runremotetest(ff *common.FuncFile, in *common.InputMap, expect
 	}
 	var stdout bytes.Buffer
 
-	if err := client.CallFN(t.provider, t.remote, ff.Path, stdin, &stdout, "", envVars, "application/json", false); err != nil {
```
riconnon: We should probably add a test for invoking functions here.

client/invoke.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/go-openapi/runtime/logger"
+)
+
+func Invoke(provider provider.Provider, appName, fnID string, content io.Reader, output io.Writer, method string, env []string, contentType string, includeCallID bool) error {
+
+	u, err := provider.CallURL(appName)
+	if err != nil {
+		return err
+	}
+	u.Path = "invoke/" + fnID
```
riconnon: I see that there are complications with using the swagger generated client due to provider-specifics, etc.

Could we, however, still support invoke in fn_go and just call that in the CLI?

client/invoke.go
```diff
@@ -0,0 +1,109 @@
...
+
+	if logger.DebugEnabled() {
+		b, err := httputil.DumpRequestOut(req, content != nil)
+		if err != nil {
+			return err
+		}
+		fmt.Printf(string(b) + "\n")
+	}
+
+	resp, err := httpClient.Do(req)
```
riconnon: You seem to use `resp` here before checking `err`. I think there's a risk of a segfault.

mantree: moved the err check

client/invoke.go
mantree: we could move these things that look like client extensions down into fn_go i guess, but I wonder if there was a reason that wasn't done before though...

commands/test.go
mantree: tested

client/invoke.go
rdallman: `/r/` wasn't ever intended to be a first class part of the API (though in practice it kinda was), since users probably want custom routes w/ e.g. dns they'd configure. but in any case, invoke a function now seems like it is part of the api proper

rdallman: as such, we may want to version it instead of `/invoke`... we should maybe clean those up, there are a few endpoints down there

client/invoke.go
```diff
@@ -0,0 +1,105 @@
...
+	if logger.DebugEnabled() {
+		b, err := httputil.DumpResponse(resp, true)
+		if err != nil {
+			return err
+		}
+		fmt.Printf(string(b) + "\n")
+	}
+
+	// for sync calls
+	if call_id, found := resp.Header[FN_CALL_ID]; found {
```
rdallman: `callID` is golint approved.
constants also should be of camel case form and not all caps, snake cased, i.e. `FnCallID` (this convention is weird but it's the golint way)

mantree: This is call_id due to a name clash with the callID struct used later. I'm a bit unwilling to change this stuff right now as it's used by the call route code. When we delete that, we can refactor this up nicely.

PR #387, opened 2018-08-17
Generate request id for each oracle provider request.

JadeRedworth:
https://github.com/fnproject/fn_go/pull/16 must be merged first.

Set the request-id on the context to be used for each oracle provider request.


objects/app/apps.go
```diff
@@ -12,11 +12,12 @@ import (
 
 	"github.com/fnproject/cli/client"
 	"github.com/fnproject/cli/common"
-	fnclient "github.com/fnproject/fn_go/client"
+	fnclient "github.com/fnproject/fn_go/client"
```
msgodf: There's some trailing whitespace here.

main.go
```diff
@@ -17,6 +20,16 @@ import (
 	"github.com/urfave/cli"
 )
 
+func getRequestID() string {
```
msgodf: Is this function based on a standard?

mantree:
Is this not already happening in the fn_go layer due to this:

https://github.com/fnproject/fn_go/commit/11f1fff5ffb3c29454a023bc385817435e50b8e2

JadeRedworth:
@mantree It is, but now we've added the request ID to the context from the CLI for all requests which allows the user to see the request-id.

zootalures:
@msgodf this is the end client ID so can be an opaque string I think this is the normal way of generating 

@mantree as jade says request ID should pass to API from context, sorry didn’t think when I reviewed originally

mantree:
Ah ok, I was only thinking to make it visible when DEBUG=1.

Issue #312, opened 2018-06-19
Error message for `fn run`, `fn build`, etc. when Docker isn't running could be more informative


msgodf:
The current error message is "ERROR: could not check Docker version: exit status 1"

This would be more informative if it suggested checking whether Docker is running.

Azaradel:
I'd like to take this on me.

PR #401, opened 2018-08-30
Change error content to be more informative

Azaradel:
Issue: #312

I changed the error content for 'run' and 'build' command when docker is not running to be more informative for user.


Azaradel:
Done!

Issue #393, opened 2018-08-24
fn deploy --local now expects registry to be set


zootalures:
It used to be the case that `fn deploy --local` ignored the absence of a configured registry - as of recently this is no longer the case

```bash
 fn deploy --local  --app foo                                                    1
Deploying sac-java-t-fn to app: foo
Bumped to version 0.0.6
Building image sac-java-t-fn:0.0.6
Updating function sac-java-t-fn using image sac-java-t-fn:0.0.6...

Fn: image name must have a dockerhub owner or private registry. Be sure to set FN_REGISTRY env var, pass in --registry or configure your context file

See 'fn <command> --help' for more information. Client version: 0.4.150

```

rikgibson:
I wonder if it's time, for a new local Fn install, to populate the initial `default` context with a simple `local` value for the `REGISTRY` field, and have the `default` context selected OOTB. Then we're usable for local development right off the bat without having to document any further config steps or populating env vars.

Having said that, even the tutorials say to populate FN_REGISTRY with a value of `fndemouser` when just doing local deployments.

rikgibson:
Also pondering whether `--local` might be better rephrased as `--no-push`?

rdallman:
we could delay it until a user types `fn deploy` for real without `--local` but it's not a great ux since it will just blow up later. fwiw i was wrestling with this the other day, as well, and was not a fan of dealing with the registry business in any capacity, and it took me a minute to figure it out (I have old man river I've worked on this too long bias, but still). i've always found FN_REGISTRY really strange. from what i can tell it exists solely so that people can clone tutorials and push `fn deploy` with _their_ fn registry set instead of having to rename the images all over, which i get, i'm just having a hard time coping with it being the preferred direction to push users. many users on a team for instance will be pushing to their company's docker registry and can code all the names in. i've personally never used it either and preferred to hardcode `fnproject/xxx` or `rdallman/xxx`. the demo clone use case while important seems not the primary use case, at least. there's my $.02 (calling for murder), sorry if tangential

zootalures:
This issue was fundamentally fixed by https://github.com/fnproject/cli/pull/403

PR #403, opened 2018-08-31
Change ValidateImageName function to ignore registry field in the image name.

JadeRedworth:
`registry` does not need to be set when validating the image name. Instead, the user should be warned that registry is missing when it is needed.

Issue [#393](https://github.com/fnproject/cli/issues/393) fn deploy --local now expects registry to be set

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -485,7 +485,7 @@ func DockerPush(ff *FuncFile) error {
 
 // DockerPush pushes to docker registry.
 func DockerPushV20180708(ff *FuncFileV20180708) error {
-	_, err := ValidateImageName(ff.ImageNameV20180708())
+	err := ValidateImageName(ff.ImageNameV20180708())
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


msgodf:
This changes the behaviour of `CreateFn`, so that a function can't be created unless a registry is specified. Which stops us creating functions against a local fn server where the images are on the local Docker - should we do this?

JadeRedworth:
@msgodf I have changed it to validate the image name for just a tag create/update and registry for docker push.

common/common.go
```diff
@@ -499,14 +499,23 @@ func DockerPushV20180708(ff *FuncFileV20180708) error {
...
 	parts := strings.Split(n, "/")
+	fmt.Println("Parts: ", parts)
 	if len(parts) < 2 {
 		return errors.New("image name must have a dockerhub owner or private registry. Be sure to set FN_REGISTRY env var, pass in --registry or configure your context file")
 
 	}
+	return ValidateTagImageName(n)
+}
+
+// ValidateTagImageName validates that the full image name (REGISTRY/name:tag) is allowed for push
```
msgodf: Should this comment read `(name:tag)`?

PR #404, opened 2018-08-31
Removes routes and migrates everything to V2 APIs

mantree:
This change leaves the 'call' command as a directive to users to
migrate their routes over to functions.


riconnon:
I'm not a fan of deleting tests we can't make pass... seems like the classic haste vs speed thing.
What do we think is required to write a test for migrate that doesn't depend on the old-style init?

JadeRedworth:
@riconnon We just need to rewrite it the way that the init command itself does but pointing to a V1FuncFile. I'm happy to do it before the end of the day, I was just under the impression this shouldn't be blocked any further.

riconnon:
@JadeRedworth I guess we can keep this as an option if someone else is desperate to open an MR for something that depends on this, but I don't think the work to remove routes from the server has begun yet, so probably you have time to get this done.

denismakogon:
Can we also get https://github.com/fnproject/cli/issues/411 fixed as part of this PR?

test/cli_misc_test.go
```diff
@@ -88,19 +88,55 @@ func TestSettingMillisWorks(t *testing.T) {
...
+	h.WithMinimalFunctionSource()
+	h.FileAppend("func.yaml", "memory: 100\nschema_version: 20180708\n")
 	h.Fn("--verbose", "deploy", "--app", appName, "--local").AssertSuccess()
-	h.Fn("call", appName, funcName).AssertSuccess()
-	inspectRes := h.Fn("inspect", "route", appName, funcName)
+	h.Fn("invoke", appName, funcName).AssertSuccess()
+
+	inspectRes := h.Fn("inspect", "fn", appName, funcName)
 	inspectRes.AssertSuccess()
-	if !strings.Contains(inspectRes.Stdout, `"cpus": "50m"`) {
```
denismakogon: i wounder if this is a regression

JadeRedworth:
@denismakogon I don't think this PR should be blocked by this issue #411. I will look into this issue now!

denismakogon:
@JadeRedworth routes from V1 is the only one way to run async functions at this moment. We have lots of demos that require async functions. But at this moment, fns and triggers are not capable to provide the same features. So, we need to have async calls in the first place before completely switching to V2.

mantree:
@denismakogon async calls are not going to be supported for the time being on fns and triggers.

test/cli_crud_test.go
```diff
@@ -83,97 +66,3 @@ func TestSimpleFnTriggerUpdateCycle(t *testing.T) {
 	h.Fn("update", "trigger", appName1, funcName1, "mytrigger", "--annotation", `"val1='["val2"]'"`).AssertSuccess()
 	h.Fn("config", "trigger", appName1, funcName1, "mytrigger", "confA", "valB").AssertSuccess()
 }
-func TestRemovingRouteAnnotation(t *testing.T) {
```
riconnon: We should probably migrate this test instead of deleting it. We don't test annotation removal on any other object.

test/cli_crud_test.go
```diff
@@ -83,97 +66,3 @@ func TestSimpleFnTriggerUpdateCycle(t *testing.T) {
...
-	h := testharness.Create(t)
-	defer h.Cleanup()
-	appName1 := h.NewAppName()
-	h.Fn("create", "route", appName1, "myroute", "foo/duffimage:0.0.1", "--annotation", "test=1").AssertSuccess()
-	h.Fn("inspect", "route", appName1, "myroute").AssertSuccess().AssertStdoutContainsJSON([]string{"annotations", "test"}, 1.0)
-	h.Fn("update", "route", appName1, "myroute", "foo/duffimage:0.0.1", "--annotation", `test=""`).AssertSuccess()
-	h.Fn("inspect", "route", appName1, "myroute").AssertSuccess().AssertStdoutMissingJSONPath([]string{"annotations", "test"})
-}
-
-func TestInvalidAnnotationValue(t *testing.T) {
```
riconnon: This tests odd behaviour in route annotations, but it might be worth migrating it to
a) test functions
b) test that the object is NOT created

test/cli_crud_test.go
```diff
@@ -83,97 +66,3 @@ func TestSimpleFnTriggerUpdateCycle(t *testing.T) {
...
-	h := testharness.Create(t)
-	defer h.Cleanup()
-	appName1 := h.NewAppName()
-
-	// The route should still be created, but without the invalid annotation
-	h.Fn("create", "route", appName1, "myroute", "foo/duffimage:0.0.1", "--annotation", "test=value").AssertSuccess().AssertStderrContains("Unable to parse annotation value 'value'. Annotations values must be valid JSON strings.")
-	h.Fn("inspect", "route", appName1, "myroute").AssertSuccess().AssertStdoutMissingJSONPath([]string{"annotations", "test"})
-}
-
-func TestRouteUpdateValues(t *testing.T) {
```
riconnon: We don't test updating values on any other object. Let's migrate this test to fns.

commands/call.go
```diff
@@ -1,62 +1,30 @@
...
-		},
-		ArgsUsage:   "<app-name> </path>",
-		Flags:       route.CallFnFlags,
+		Name:        "call",
+		Usage:       "\tPrompts for migration.",
+		Aliases:     []string{"cl"},
+		ArgsUsage:   "",
 		Category:    "DEVELOPMENT COMMANDS",
-		Description: "This command calls explicitly a function.",
+		Description: "This command no longer executes a function, instead it prompts users to migrate their routes to fn/triggers.",
```
rdallman: maybe be nice to see 'you should use invoke' or link to details about the migration here

rdallman: nvm, it does do that. just not here i guess

commands/deploy.go
```diff
@@ -480,30 +399,37 @@ func expandEnvConfig(configs map[string]string) map[string]string {
...
+	app, err := apps.GetAppByName(p.clientV2, appf.Name)
 	if err != nil {
-		postParams := clientApps.NewPostAppsParams() //XXX switch to put when v2.0 Fn
-		postParams.Body = &models.AppWrapper{
-			App: &models.App{
+		l := log.New(os.Stderr, "", 0)
+		l.Printf("%T", err)
+		switch err.(type) {
+		case apps.NameNotFoundError:
+			l.Println("THIS IS HAPPENING")
```
rdallman: this shouldn't happen :(

rdallman: few prints in here to clean up i think

commands/get.go
```diff
@@ -13,7 +13,7 @@ func GetCommand() cli.Command {
 		Category:    "MANAGEMENT COMMANDS",
 		Hidden:      false,
 		ArgsUsage:   "<subcommand>",
-		Description: "This commands gets a 'call', 'configuaration' or 'log' to retrieve information for an object ('app', 'route' or 'function').",
+		Description: "This commands gets a 'call', 'configuaration' or 'log' to retrieve information for an object ('app' or 'function').",
```
rdallman: s/configuaration/configuration/

objects/call/calls.go
```diff
@@ -21,12 +23,12 @@ type callsCmd struct {
...
 
 // getMarshalableCall returns a call struct that we can marshal to JSON and output
-func getMarshalableCall(call *models.Call) interface{} {
+func getMarshalableCall(call *modelsv2.Call) interface{} {
 	if call.Error != "" {
 		return struct {
 			ID          string          `json:"id"`
 			AppID       string          `json:"appId"`
-			Path        string          `json:"path"`
+			FnID        string          `json:"fnId"`
```
rdallman: all of our apis in fn are snake case... but i'll punt ;)

riconnon: `appId` was already like that and I don't want to change this now

test/cli_misc_test.go
rdallman: good catch, we have removed it from API in fns though

PR #409, opened 2018-09-06
Make BuildFunc context-free

denismakogon:
It's not a secret that we've put a lot of efforts to CLI to make it what it is right now.
At this moment, CLI contains a lot of useful logic that may be reused while developing another tool which may require the same set of features.

The problem is we pass down the CLI context even if that is used for 1 flag value (consider here `--verbose`) which makes the whole CLI's internal logic not reusable what forces developers to solve the same problems.

In this change, I was trying to get rid of the CLI context by substituting one with parameter(s) that is required by the particular method (consider here a logic behind `fn deploy`).

### DIFF
--- a/config/version.go
+++ b/config/version.go
@@ -11,7 +11,6 @@ import (
 // Version of Fn CLI
 var Version = "0.4.154"
 
-
 func GetVersion(versionType string) string {
 	base := "https://github.com/fnproject/cli/releases"
 	url := ""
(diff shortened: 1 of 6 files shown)


config/version.go
```diff
@@ -11,7 +11,6 @@ import (
 // Version of Fn CLI
 var Version = "0.4.154"
 
-
```
denismakogon: picked up by `go fmt ./...`

PR #413, opened 2018-09-11
Update init command

JadeRedworth:
Remove the creation of v1 func files with the init command:

`fn init --runtime go my-func` will create a v2 funcfile without trigger:
```
schema_version: 20180708
name: hello2
version: 0.0.1
entrypoint: ./func
format: json
``` 
`fn init --runtime go  --trigger my-func` will create a v2 funcfile with `http` trigger:
```
schema_version: 20180708
name: hello3
version: 0.0.1
entrypoint: ./func
format: json
triggers:
- name: hello3-trigger
  type: http
  source: /hello3-trigger
```

### DIFF
--- a/commands/init.go
+++ b/commands/init.go
@@ -33,18 +33,15 @@ import (
 	"github.com/fnproject/cli/common"
 	"github.com/fnproject/cli/langs"
 	function "github.com/fnproject/cli/objects/fn"
-	"github.com/fnproject/cli/objects/route"
-	models "github.com/fnproject/fn_go/models"
 	modelsV2 "github.com/fnproject/fn_go/modelsv2"
 	"github.com/urfave/cli"
 )
 
 type initFnCmd struct {
-	force       bool
-	triggerType string
-	wd          string
-	ff          *common.FuncFile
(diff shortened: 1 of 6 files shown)


commands/init.go
```diff
@@ -601,11 +432,11 @@ func (a *initFnCmd) BuildFuncFile(c *cli.Context, path string) error {
 		}
 
 		if runtime == "" {
-			runtime = helper.Runtime()
+			a.ff.Runtime = helper.Runtime()
+		} else {
+			helper.Runtime()
```
mantree: this doesn't look right. Should this block actually get changed?

commands/init.go
```diff
@@ -33,18 +33,15 @@ import (
...
 )
 
 type initFnCmd struct {
-	force       bool
-	triggerType string
-	wd          string
-	ff          *common.FuncFile
-	ffV20180708 *common.FuncFileV20180708
+	force   bool
+	trigger string
```
mantree: triggerType I think is still the right name here

zootalures:
looks like the ruby tests are broken

Issue #411, opened 2018-09-09
"fn migrate" config migration error


denismakogon:
func.yml:
```yaml
name: bucket-daemon
version: 0.0.5
runtime: go
entrypoint: ./func
type: async
memory: 256
format: json
timeout: 3600
path: /bucket-daemon
config:
  BACKOFF_TIME: 5
  NEXT_FUNC: "/segments-assembler"
idle_timeout: 500
```

```
$ fn migrate
The func.yaml is not valid. Please see errors: 
- config: Invalid type. Expected: integer, given: object

Fn: Please update your func.yaml to satisfy the schema.

See 'fn <command> --help' for more information. Client version: 0.4.155

```

```
$ fn version
Client version is latest version: 0.4.155
Server version:  0.3.548

```

PR #414, opened 2018-09-11
Update config in json schema to 'object'

JadeRedworth:
Fix: [#411 "fn migrate" config migration error](https://github.com/fnproject/cli/issues/411)


mantree:
Can we get a check that this fixes the reported problem?

PR #415, opened 2018-09-11
remove welcome message

rdallman:
this supposedly runs on new installs, but I:

* don't think it's very useful (not having a changelog withstanding)
* it prints to stdout, which messes up output, which can mess up scripts, etc
* doesn't seem to run as was intended, and doesn't seem like a very good place to be
in case e.g. somebody is piping invoke output to a script... and I can't think
of any place it belongs (we tried to do some cutesy ascii art in `fn init`
once, that was also annoying). open to ideas for better places to put this, i
just can't think of any


rikgibson:
I agree the original change maybe was a bit quick-and-dirty but the overall intent of somehow informing the user that "hey you've just picked a new version and some stuff has changed" isn't unreasonable I think.  Having better release notes / changelog is also not unreasonable, which was really the idea here - "Things have changed - click this link to see what..."

riconnon:
I'm in the "this is unnecessary" camp, but if we're keeping it absolutely needs to be in stderr not stdout.

PR #425, opened 2018-09-17
fixes default url to not have /v1

### LINKED ISSUE #424: FN CLI 0.5.1 not working
i have pointed to the correct FN_API_URL.

I get the below error for all commands

Fn list functions {app name}

Fn: [GET /apps][404] ListApps default  &{Fields: Message:}

See 'fn <command> --help' for more information. Client version: 0.5.1

rdallman:
it seems we append this now, and including /v2 means we get /v2/v2, and
even xzhibit doesn't like /v2 that much. this gets rid of all the references
to /v1 that were laying around for the api url. guess we forgot to test this,
oops.

also updated the deps with just `dep ensure` -- this means they were stale, I
didn't type update, so this should fix the mismatch there...

fixes #424


rdallman:
kind of in a rush to fix master here, i tested this working but am not sure about tests passing / should we parse this out in the future / etc etc...

rdallman:
thanks @kojustin

Issue #383, opened 2018-08-14
Deprecate and remove `fn build` and `fn test`


mjg123:
At the moment the CLI does two distinct things:

  - builds images
  - calls the API in order to deploy functions

As far as "calls the API" goes, that's fine.  Makes sense that we provide a tool for doing that.  But the justification for "Builds images" is significantly weaker.  We couldn't possibly make a better image building tool than docker itself, and we're spending a lot of time and effort trying to.

### Proposal:

  - `fn init` always generates a Dockerfile
  - `fn build` becomes a synonym for `docker build`
  - `fn test` removed
  - `func.yaml` is purely a deployment descriptor. The following keys which relate to building images on the host are deprecated and removed:
    - build, build_image, cmd, content_type, entrypoint, expects, format, headers, run_image, runtime, tests

### Work needed:

  - init-images: #369 
  - reduction in number of hard-coded runtimes: #380 
  - supported init-images should produce dockerfiles and `runtime: docker` in the boilerplate
  - remove all LangHelper code from CLI
  - deprecate build-related keys from func.yaml
  - remove all build code from CLI

shaunsmith:
I agree with the goal of extracting all language specific logic from the CLI and relying on the generation of an appropriate Dockerfile.  See #369 for comments on `init`.

+1 on removal of `fn test`.  I don't think this is being used.

Won't `func.yaml` `runtime:` always be `docker` and therefore unnecessary?

zootalures:
‘runtime’ is only used for build I think  (to generate the Dockerfile) so yes, becomes irrelevant

mjg123:
I think we can also remove `--entrypoint` and `--cmd` from `fn init` as well.

mjg123:
Tagging in @mantree @JadeRedworth @riconnon and @rdallman for their thoughts too.

mantree:
This sounds like a great idea. Less is more, and what little there is should be tightly focused.

:+1:

denismakogon:
I agree that `fn build` and `fn test` commands are quite useless.

However, if we gonna drop `fn test` we need to provide something else. FDK-Java started with their own integration with JUnit to let developers test their functions.
I did the same for FDK-python: https://github.com/fnproject/fdk-python/pull/47

So, we need to do the following:

1. Get rid of `fn test`, it's useless for something more complex than hello-world.
2. Write a boilerplate for each FDK that would represent a test suite, here's an example of an extended boilerplate:

```python
import fdk
import ujson

[...]

PR #429, opened 2018-09-18
rm build command (#383)

josephholsten:
The absolute minimum that could possibly qualify as #383.


rdallman:
there's not a strong argument to get rid of this and #383 was in actuality fleshing out init behavior mostly (the title is misleading), this command is still useful to mask the docker commands from the user imo and there's a case for building functions locally without bumping for things like testing them. thanks for the PR, i would close but maybe @mjg123 and @rikgibson should have a look here first

riconnon:
We definitely can't do this until all supported runtime options output a `Dockerfile` from `fn init`

rikgibson:
I agree with @rdallman - we're not going down the #383 path, certainly not anytime soon. Thanks again for the PR but better closing this off at this time.

PR #432, opened 2018-09-24
Default to http-stream format for new functions

riconnon:
Should probably wait for a few more FDKs, but this is here when we want it.

### DIFF
--- a/client/invoke.go
+++ b/client/invoke.go
@@ -16,7 +16,7 @@ import (
 )
 
 const (
-	FN_CALL_ID             = "Fn_call_id"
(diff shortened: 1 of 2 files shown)


riconnon:
@rikgibson what about `--runtime docker`

riconnon:
actually... looks like that creates a funcfile with no `format`

rikgibson:
` --runtime value                Choose an existing runtime - go, java8, java9, java, node, python, python3.6, python, python3.7, ruby, kotlin`

`docker` is not a supported value at this time - if the CLI is still accepting it as such then that should be removed too.

Issue #407, opened 2018-09-04
`fn update context` should support deleting context entries


msgodf:
The only way to remove something from the context at the moment is to edit the file.

It would be nice if `fn update context` either removed the line if no value is provided, or if a `--delete` flag is passed in.

PR #434, opened 2018-09-25
Add --delete flag to update context command

JadeRedworth:
When updating the current context, the '--delete' flag will indicate users want to remove the key/value pair from the current context file.

Fix: [#407 `fn update context` should support deleting context entries](https://github.com/fnproject/cli/issues/407)


JadeRedworth:
@riconnon Thanks for creating [#437](https://github.com/fnproject/cli/issues/437) 👍 

I have fixed the issue with 'point 2' we now check if 'value' is empty and error.

riconnon:
Could we add tests for `--delete` and the new empty value error?

PR #440, opened 2018-09-28
Remove run and test commands.

riconnon:
Currently these two commands are both broken using http-stream
functions. Run should almost certainly just rest in peace, possibly we
want to bring back integration testing of functions at a later date, but
this would now have to be run through an Fn server so it would look
nothing like it currently does.


riconnon:
Before merging this someone should test invoke on windows.

riconnon:
@tteggel did you test `fn invoke` using this branch on windows?

PR #448, opened 2018-10-02
Making the docker error message for "fn build" more informative

### LINKED ISSUE #312: Error message for `fn run`, `fn build`, etc. when Docker isn't running could be more informative

shreyagarge:
Added an additional check when "fn build" fails because the docker is not running.
Checks the exit status from docker version command and prints a message asking to check if the docker is running - to make it more informative.

fixes #312

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -297,14 +297,14 @@ func dockerVersionCheck() error {
 		if errString == "exit status 1" {
 			return fmt.Errorf("Cannot connect to the docker daemon. Is the docker daemon running?: %v", err)
 		}
-		return fmt.Errorf("could not check Docker version test1: %v", err)
+		return fmt.Errorf("could not check Docker version: %v", err)
 	}
 	// dev / test builds append '-ce', trim this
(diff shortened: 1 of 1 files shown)


common/common.go
```diff
@@ -290,8 +290,13 @@ func RunBuild(verbose bool, dir, imageName, dockerfile string, buildArgs []strin
...
 func dockerVersionCheck() error {
-	out, err := exec.Command("docker", "version", "--format", "{{.Server.Version}}").Output()
-	if err != nil {
+	out, err := exec.Command("docker", "version", "--format", "{{.Server.Version}}").Output() //executes the "docker version" command and gets the output and error code
+	if err != nil {                                                                           //returns exit status 1 err if docker is not running
+		errString := err.Error() //converts the err to string
+		//fmt.Printf(errString)
+		if errString == "exit status 1" {
+			return fmt.Errorf("Cannot connect to the docker daemon. Is the docker daemon running?: %v", err)
+		}
```
rdallman: I'm not sure we need this after https://github.com/fnproject/cli/pull/441/files ?

PR #450, opened 2018-10-04
Check for verbose flag in build failure error message

### LINKED ISSUE #307: `fn build` failure suggests using `--verbose` flag even when it is already specified
Running

`fn --verbose build`

gives the output

`Error during build. Run with `--verbose` flag to see what went wrong. eg: `fn --verbose CMD``

when the build fails. It shouldn't suggest it when it is already being specified.

shreyagarge:
Added a check if verbose flag is set when fn builds fails.

Fixes #307


tteggel:
Shreya is an Oracle employee. Merging.

Issue #449, opened 2018-10-03
Show App , Function and Trigger IDs more prominantly in CLI 


zootalures:
Fn list operations should include IDs in the tabluar responses  - these are necessary to do a bunch of operations since V2 

right now you need do do an inspect, then copy and paste from a JSON body

denismakogon:
Should we just allow to dump out the JSON payload for `list` command? Or do both?

delabassee:
What about adding support for human parsable alias for FunctionId?

zootalures:
Right now I just want the IDs in list operations I think - aliasing is complex and I don't want to pollute the domain model yet - maybe later.

zootalures:
i.e.  with v2 model we have operations that depend on users knowing IDs and they are hard to find now.

PR #453, opened 2018-10-08
added ID info to fn list output in app, funcs & triggers

nishalad95:
fix: https://github.com/fnproject/cli/issues/449

### DIFF
--- a/testharness/harness.go
+++ b/testharness/harness.go
@@ -403,6 +403,7 @@ runtime: docker
 //NewFuncName creates a valid function name and registers it for deletion
 func (h *CLIHarness) NewFuncName(appName string) string {
 	funcName := randString(8)
+	//funcID := randString(26)
 	h.funcRefs = append(h.funcRefs, funcRef{appName, funcName})
 	return funcName
 }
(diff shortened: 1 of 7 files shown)


zootalures:
Looks like we have some (perhaps over-sensitive) tests that don't like the new format.

testharness/harness.go
```diff
@@ -403,13 +403,15 @@ runtime: docker
 //NewFuncName creates a valid function name and registers it for deletion
 func (h *CLIHarness) NewFuncName(appName string) string {
 	funcName := randString(8)
+	//funcID := randString(26)
```
zootalures: Can you delete these

Otherwise gtg

testharness/harness.go
```diff
@@ -403,13 +403,15 @@ runtime: docker
...
 	h.funcRefs = append(h.funcRefs, funcRef{appName, funcName})
 	return funcName
 }
 
 //NewTriggerName creates a valid trigger name and registers it for deletioneanup
 func (h *CLIHarness) NewTriggerName(appName, funcName string) string {
 	triggerName := randString(8)
+	//triggerID := randString(26)
```
zootalures: Delete this

nishalad95:
Sure, ...I was in the process of making a test to check the IDs, but if it's not needed I'll make the changes and commit

Issue #455, opened 2018-10-15
Creating a function does not default to "http-stream" format


JadeRedworth:
When creating functions and not supplying a `--format` the CLI will fail with invalid format. The format should default to creating functions in the right format `http-stream`.
 
```
$ fn create func justin-app hello-world i386/hello-world:linux --memory=128 --timeout=120
Fn: Invalid format on Fn
See 'fn <command> --help' for more information. Client version: 0.5.15
```

rikgibson:
Tested on latest CLI - 0.5.29 - against local Fn and remote Oracle Functions server and this is no longer an issue, so closing.

PR #456, opened 2018-10-15
Add default format 'http-stream' when creating a function.

JadeRedworth:
Fix: [#455 Creating a function does not default to "http-stream" format](https://github.com/fnproject/cli/issues/455)

### DIFF
--- a/commands/init.go
+++ b/commands/init.go
@@ -102,10 +102,6 @@ func initFlags(a *initFnCmd) []cli.Flag {
 			Name:  "headers",
 			Usage: "Function response headers",
 		},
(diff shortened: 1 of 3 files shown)


rdallman:
+1 at this point we can probably say 'use old cli' - am in the process of removing on server side this week

rdallman:
@JadeRedworth I think we can close this and make a PR to remove format from CLI altogether? atm API will just ignore it, so it's fine, but we can get rid of it now.

rdallman:
feel free to re-open / update* however you'd like to do things...

Issue #462, opened 2018-10-19
Fn cli "fn update app config" doesn't work


JadeRedworth:
Updating an apps config with: `fn update app --config KEY=VALUE <app-name>` does not work. You can update the config with: `fn config app <app-name> KEY VALUE`.

abhirockzz:
@JadeRedworth  The same issue applies to `syslog-url` as well i.e. if there is an existing value `fn update app --syslog-url=newvalue myapp` will not work. I guess the fix is as simple as `if len(c.String("syslog-url")) > 0`

abhirockzz:
Filed an issue - https://github.com/fnproject/cli/issues/464

JadeRedworth:
@abhirockzz ok I will include a fix in the same PR then!

PR #463, opened 2018-10-19
Fix update app by using --config and --syslog-url flags.

JadeRedworth:
When updating App with --config or --syslog-url flag, the value is not saved.

Fix: [#462 Fn cli "fn update app config" doesn't work]( https://github.com/fnproject/cli/issues/462)
Fix: [#464 Fn update app --syslog-url" does not work](https://github.com/fnproject/cli/issues/464)

### DIFF
--- a/objects/app/apps.go
+++ b/objects/app/apps.go
@@ -31,7 +31,7 @@ func printApps(c *cli.Context, apps []*modelsv2.App) error {
 		for _, app := range apps {
 			a := struct {
 				Name string `json:"name"`
-				ID string `json:"id"`
+				ID   string `json:"id"`
 			}{app.Name,
 				app.ID,
 			}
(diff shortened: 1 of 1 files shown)


kojustin:
yessssss!!!!!!!!

objects/app/apps.go
```diff
@@ -92,10 +92,10 @@ func (a *appsCmd) list(c *cli.Context) error {
 }
 
 func appWithFlags(c *cli.Context, app *modelsv2.App) {
-	if app.SyslogURL == "" {
+	if len(c.String("syslog-url")) > 0 {
 		app.SyslogURL = c.String("syslog-url")
 	}
-	if len(app.Config) == 0 {
+	if len(c.StringSlice("config")) > 0 {
```
kojustin: does string slice allow specifying this multiple times? because that would be awesome.

kojustin:
Should we add tests?

PR #467, opened 2018-10-22
BinTray TLS fix broke HTTP proxy support in Java lang

tteggel:
The fix to the TLS config for bintray inadvertently broke http proxy support for Java. This fix reinstates this by basing the new transport config from DefaultTransport instead of a totally new config.

### DIFF
--- a/langs/java.go
+++ b/langs/java.go
@@ -206,10 +206,17 @@ func (lh *JavaLangHelper) getFDKAPIVersion() (string, error) {
 	}
 
 	// nishalad95: bin tray TLS certs cause verification issues on OSX, skip TLS verification
-	transCfg := &http.Transport{
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, // ignore expired SSL certificates
+	defaultTransport := http.DefaultTransport.(*http.Transport)
+	noVerifyTransport := &http.Transport{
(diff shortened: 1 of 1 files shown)


langs/java.go
```diff
@@ -206,9 +206,8 @@ func (lh *JavaLangHelper) getFDKAPIVersion() (string, error) {
 	}
 
 	// nishalad95: bin tray TLS certs cause verification issues on OSX, skip TLS verification
-	transCfg := &http.Transport{
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, // ignore expired SSL certificates
-	}
+	transCfg := http.DefaultTransport
```
zootalures: `http.DefaultTransport` is an interface so this will modify by reference( modifying this will modify the global behaviour ) 
it's messy but it's better to assign a copy: 
```
	var transCfg http.Transport
	transCfg = *(http.DefaultTransport.(*http.Transport))
	transCfg.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
	client := &http.Client{Transport: &transCfg}

```

tteggel: This will result in copying the `idleMu     sync.Mutex` from DefaultTransport which I think is not a good idea? Copying in values from DefaultTransport seems the way to go.

PR #471, opened 2018-10-25
remove environment expansion from config

zootalures:
The current EnvExpand stuff removes/mutates  config values that naturally contain dollars - this is completely unexpected and has caused a few issues with config values being re-written. 

I propose to just remove that and make the config the config - if you want to env expansion that's what shells are for.

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -457,13 +457,13 @@ func stringToSlice(in string) string {
 	return buffer.String()
 }
 
-// ExtractEnvConfig extract env vars from configuration.
-func ExtractEnvConfig(configs []string) map[string]string {
+// ExtractConfig parses key-value configuration into a map
+func ExtractConfig(configs []string) map[string]string {
 	c := make(map[string]string)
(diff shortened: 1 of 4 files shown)


common/common.go
```diff
@@ -457,13 +457,13 @@ func stringToSlice(in string) string {
...
-// ExtractEnvConfig extract env vars from configuration.
-func ExtractEnvConfig(configs []string) map[string]string {
+// ExtractConfig parses key-value configuration into a map
+func ExtractConfig(configs []string) map[string]string {
 	c := make(map[string]string)
 	for _, v := range configs {
 		kv := strings.SplitN(v, "=", 2)
 		if len(kv) == 2 {
-			c[kv[0]] = os.ExpandEnv(kv[1])
+			c[kv[0]] =kv[1]
```
rdallman: add fmt on save to your editor already ;)

zootalures: sorry - had it fn , just not CLI

PR #472, opened 2018-10-25
Remove format-dependent tests

zootalures:
Updates the integ tests to used a pre-baked golang function, removes checks on format params. 
This makes the CLI tests runnable after we get rid of formats #456  can finish off the rest


langs/kotlin.go
```diff
@@ -189,10 +189,17 @@ func (lh *KotlinLangHelper) getFDKAPIVersion() (string, error) {
 	}
 
 	// nishalad95: bin tray TLS certs cause verification issues on OSX, skip TLS verification
-	transCfg := &http.Transport{
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, // ignore expired SSL certificates
+	defaultTransport := http.DefaultTransport.(*http.Transport)
```
zootalures: Noticed that we weren't doing this hack for kotlin  but were for java

langs/node.go
```diff
@@ -51,7 +51,7 @@ const packageJsonContent = `{
 	"author": "",
 	"license": "Apache-2.0",
 	"dependencies": {
-		"@fnproject/fdk": "0.x"
+		"@fnproject/fdk": ">=0.0.11"
```
zootalures: also randomly I saw a build pick up a pre-http-stream version  might just have been docker cache but added this to be sure.

rdallman:
@zootalures spent at least an hour trying to debug this, I'm convinced you managed to get these tests passing locally with some files that may not have been checked in to the remote branch. building `simplefunc` works once adding `vendor/` but other tests that are building functions fail. I'm going to bail on this one for today, sorry that I couldn't manage to fix this up.

zootalures:
Test function was missing a vendor dir - sorry was excluded from checking

Issue #485, opened 2018-11-20
Speedup python builds


denismakogon:
For now, every python build takes at least `1m31.005s` which is quite a lot.
Here's why, we `ADD . /function/` before checking if `requirements.txt` exists, in Python, you can install dependencies without having bits of code.

Let's add `requirements.txt` before adding the rest files from the directory.
With this improvement, both `deploy` and `build` will take not longer than `0m16.046s`

PR #486, opened 2018-11-20
A new boilerplate for a new Python FDK

denismakogon:
A new FDK Python requires a new boilerplate.

Closes: #485


denismakogon:
Why this actually helps: when you change the code you don't need to rebuild the dependencies, except those cases when you add a new dependency in code, but it will also appear in `requirements.txt` and will trigger a build, so, we're safe here.

denismakogon:
Heavily depends on https://github.com/fnproject/fdk-python/pull/60.

denismakogon:
Because only 3.7.1 support importlib required by an FDK.

denismakogon:
To whom it may concern, CI will go green when a new FDK will be released.

denismakogon:
new FDK released, updating this PR in order to make sure that everything works as it suppose to be

denismakogon:
@rdallman circle is happy, i'd like to proceed with this PR. Can you please approve it?

Issue #406, opened 2018-09-04
--init-image should fail if no correct image has been provided


delabassee:
`fn init --init-image=void` just hangs forever

Issue #405, opened 2018-09-03
--init-image should fail quickly if Docker isn't running


delabassee:
Given that the new _init-image_ approach relies on Docker, the CLI should test if Docker is running. Right now, the CLI just hangs forever when Docker isn't running.

PR #488, opened 2018-11-26
simple fix for #405 and #406

delabassee:
Simple fix that uses `docker inspect` to check that Docker is running and that the init-image exists.

### DIFF
--- a/commands/init.go
+++ b/commands/init.go
@@ -315,6 +315,11 @@ func (a *initFnCmd) init(c *cli.Context) error {
 func runInitImage(initImage string, a *initFnCmd) error {
 	fmt.Println("Building from init-image: " + initImage)
 
+	cmdCheck := exec.Command("docker", "inspect", "--type=image", initImage)
+	if cmdCheck.Run() != nil {
+		return errors.New("Error, can't pull the init-image! Make sure that Docker is running and that the init-image exists.")
+	}
+
 	// Run the initImage
 	var c1ErrB bytes.Buffer
 	tarR, tarW := io.Pipe()


rdallman:
is the pull step before this? lazy and didn't see it in the diff (sorry). users shouldn't need to manually pull the init image and docker inspect will burp if the image isn't downloaded locally (which was the point I think? got that) -- docker build itself will pull the image, i'm assuming we were just relying on that, which makes sense to me. `docker info` is a quick way to check that docker is running. `docker search` is a quick way to check that an image exists on a registry. I guess here we have to balance that some users may build their own init images locally and have them and some users may be relying on them to exist on a registry to be pulled at build time (like ours, presumably, eg for newer users on their first run through). do we need to try both search and inspect to get the desired validation behavior or what exactly do we need to validate here?

delabassee:
I changed this to simply handle `docker run` potential failures.

commands/init.go
```diff
@@ -315,23 +316,34 @@ func (a *initFnCmd) init(c *cli.Context) error {
...
+		_, errStdout = io.Copy(stdout, stdoutIn)
+	}()
+
+	go func() {
+		_, errStderr = io.Copy(stderr, stderrIn)
+	}()
+
+	err = cmd.Wait()
+	if err != nil || errStdout != nil || errStderr != nil {
+		return errors.New("Error, unbale to use the init-image! Make sure that Docker is running and that the init-image exists.")
```
rdallman: think I'm grokking what's going on, but we shouldn't need the two goroutines, and we can call wait at the end regardless of whether untarStream returns an error or not? ie we need to check that the command succeeded regardless of whether the untar succeeds, and from there we can make the useful error. if the command succeeds and untar succeeds, we're good. I think the pipe was preferable to a buffer, this adds a good bit of memory overhead as images can be pretty large (few megs, at least). basically asking to move wait until after untarStream and check cmd.Wait instead of bailing if untarStream has an error (and if untarStream does have an error but cmd.Wait does not, return untarStream's error?)? not sure if that works though?

commands/init.go
```diff
@@ -315,22 +316,29 @@ func (a *initFnCmd) init(c *cli.Context) error {
...
-		return errors.New("Error running init-image")
+	stdout := io.Writer(&stdoutBuf)
+	stderr := io.Writer(&stderrBuf)
+
+	errStart := cmd.Start()
+
+	_, errStdout = io.Copy(stdout, stdoutIn)
+	_, errStderr = io.Copy(stderr, stderrIn)
+
+	errUntar := untarStream(bytes.NewReader(stdoutBuf.Bytes()))
```
rdallman: sorry for delay here. we can use the io.Pipe and not need to buffer here - this is a good idea because images are on the order of megabytes. we don't actually use stderr it seems so i'm thinking something like:

```go
tarIn, tarOut := io.Pipe()
cmd := exec.Command("...")
cmd.Stderr = ioutil.Discard // honestly, os.Stderr makes sense, we can do w/e
cmd.Stdout = tarOut
err := cmd.Start()
if err != nil {
  return err
}
err = untarStream(tarIn)
if err != nil {
  return err // can format error about init image here, add stderr if we want
}
err = cmd.Wait()
if err != nil {
  return err // can format error about init image here, add stderr if we want
}
```

this pretty much expects that untar stream will fuck up if the command fucks up. if for some reason it doesn't, then we can get the exit code out of Wait(). in the happy path we expect untar to just work, and Wait() to return a nil error. I think that this makes sense, but double check me :) wdyt?

PR #499, opened 2018-12-06
Allow syslog-url to be unset (CLI part)

hhexo:
The existing check was checking for the syslog-url option being an empty string rather than the option being set. 
This prevented the possibility of setting the syslog URL to an actual empty string (i.e. unsetting it).

### DIFF
--- a/objects/app/apps.go
+++ b/objects/app/apps.go
@@ -92,7 +92,7 @@ func (a *appsCmd) list(c *cli.Context) error {
 }
 
 func appWithFlags(c *cli.Context, app *modelsv2.App) {
-	if len(c.String("syslog-url")) > 0 {
+	if c.IsSet("syslog-url") {
 		app.SyslogURL = c.String("syslog-url")
 	}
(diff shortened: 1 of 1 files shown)


hhexo:
Hm... this is not enough, as the `fn_go` models have `omitempty` on the syslog URL field, so the string is elided again.

And the `fn_go` is generated from the swagger, where the syslog URL is not required, therefore the generator sets `omitempty`.

hhexo:
This cli change is still necessary to allow syslog-url to be unset, but it is not the whole solution: the API needs to be changed too as it currently ignores empty strings in syslog-url.

rdallman:
@hhexo hmm, maybe fn_go and the swagger doc (longstanding todo to update those bindings... maybe it's time), but I think the API response is correct https://github.com/fnproject/fn/blob/master/api/models/app.go#L170 ?

rdallman:
thanks for #1346 - need to run the client bindings script

rdallman:
just ignore me - i'll do email backwards next time

PR #513, opened 2019-01-21
Remove fn call

### LINKED ISSUE #510: [UI] Remove 'call' from the CLI Help
CLI (0.5.40) still mentions in its help `call` as one of the development commands (ex. `fn`).
Given this command is not supported anymore, it should simply be removed from the help.
`fn call` produces the appropriate "no longer supported" warning.

rikgibson:
One-liner to remove no-longer-supported `fn call`,

Fixes #510


commands/commands.go
```diff
@@ -22,7 +22,7 @@ var Commands = Cmd{
 	"build":        BuildCommand(),
 	"build-server": BuildServerCommand(),
 	"bump":         common.BumpCommand(),
-	"call":         CallCommand(),
+//	"call":         CallCommand(),
```
rdallman: can we remove the command code as well here please?

.gitignore
```diff
@@ -9,4 +9,5 @@ _vendor*/
 /cli
 data/
 .vscode/
-.idea/
\ No newline at end of file
+.idea/
+**/*.iml
```
denismakogon: worth to add to global git ignore, no?

commands/commands.go
rikgibson: duh, more haste less speed.

.gitignore
rikgibson: sure you can do that too, this does no harm though for anyone that hasn't done so.

PR #518, opened 2019-01-23
Lang helpers add custom memory remove default format

rikgibson:
Two things here:

1) Removes the DefaultFormat function from all the LangHelper variants and just bakes in "http-stream" across the board.

2) Adds a CustomMemory option to the same helper classes, to allow languages to specify a different base memory option than the default in the server runtime. Specifically, the PythonLangHelper now generates a "memory: 256" entry in `func.yaml` for python functions.  This should help python fdk functions start up quicker whilst we wait for an improved FDK.


langs/go.go
```diff
@@ -18,8 +18,8 @@ func (h *GoLangHelper) Runtime() string {
 	return h.LangStrings()[0]
 }
 
-func (h *GoLangHelper) DefaultFormat() string {
-	return "http-stream"
+func (h *GoLangHelper) CustomMemory() uint64 {
```
rdallman: not sure why most of the lang helpers don't embed a `BaseHelper` in their struct? stuff like this can just inherit and doesn't need to modify for most langs. kind of a nit I guess, most lang's methods seem to return different stuff

PR #529, opened 2019-02-06
Do not create app by default on `fn deploy`

riconnon:
Dependent on https://github.com/fnproject/cli/pull/528 in order to build/test

In response to user feedback from @shaunsmith this disables creation of apps by default with `fn deploy`

Old behaviour can be restored using `fn deploy --create-app`

This will have documentation impact!


shaunsmith:
@michael-w-williams this PR removes the default app creation and adds a new option to enable the old behaviour.

PR #531, opened 2019-02-06
install: add /usr/local/bin to PATH if not defined

blues-man:
This fixes installation on OS like CentOS 7, where the `/usr/local/bin` directory is not in the PATH by default


rdallman:
see: https://github.com/fnproject/fn/issues/1395

PR #532, opened 2019-02-07
Vendor dependencies

riconnon:
Does exactly what it says on the tin.

All the standard arguments for/against vendoring apply, but mainly the argument here is consistency with the rest of Fn


rdallman:
this appears to use dep -- i'd like to do this too, but i'd like to do https://github.com/fnproject/cli/pull/519 and then use mod to vendor the deps, if possible. it's inconsequential and we could redo this and i'm totally using this as an excuse to get someone to review the mod change ;) we should move to mod though. thanks for bringing this to attention and doing it

riconnon:
I updated this to do the following:
- Build using Go 1.11 rather than Go 1.10 in docker
- Use a consistent Go release for all builds
- Consistently build using GO111MODULE=on and GOFLAGS=-mod=vendor
- Use the same alpine release for build and run docker containers

This will fix the latest broken master build.

test.sh
```diff
@@ -21,19 +21,6 @@ export fn="${CUR_DIR}/fn"
...
-	WORK_DIR=$(mktemp -d /tmp/temp.XXXXXX)
-else
-	WORK_DIR=$(mktemp -d)
-fi
-cd $WORK_DIR
-
-# HACK: if we don't pre touch these, fn server will create these as 'root'
-mkdir data
-touch data/fn.db data/fn.mq
-
```
rdallman: did you happen to test on osx here? i'm not sure if they work on master on osx anymore with uds stuff even, just asking to keep track though.

rdallman: i see you commit msg https://github.com/fnproject/cli/pull/532/commits/8bc2e2f3f56c8da34f6ab051b76eb5485315f283 - thanks

rdallman:
merging to fix master... thank you

PR #533, opened 2019-02-08
moving app check up

rdallman:
now that we require apps to exist when deploying them, we should check whether
the app exists before running around and bumping and creating images and all
that, this makes that change. for deploy all this reduces the number of api
calls substantially, we were checking the app each function. obviously this
whole thing is not really atomic, so it can fail pretty spectacularly, but we
aren't really worried about that (and arguably this code is much safer to
delete-create race deploying to 'the wrong app' since it uses id all the way
through for deploy all)

cleaned up error handling to only create functions, triggers, and apps if they
don't exist, not just if any error happens. saves some api calls when service
is down or something, but mostly it doesn't mask any errors from doing the
original get when it's not a not found error (things like permissions this
could be useful).

reduced api calls by one when a function doesn't exist (we don't need to go
get it after creating, the api returns it with the id filled in). /sarcasticwoo

* move app yaml deploy up to app get/create at the top of deploy, reduces
calls to get and create app and cleans up the code
* fix deploy of app.yaml logic error, this test was previously passing that
should have failed, because using an app.yaml meant incidentally that the app
would get created if necessary
* add additional tests to see that app.yaml deploy actually sets the config
and an additional test that it works for both create and update of an app
* cleaned up deploy to use app object code for cleanliness (error handling
improvements)
* make tests use test fatal facilities and not log package

### DIFF
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -15,8 +15,6 @@ import (
 	trigger "github.com/fnproject/cli/objects/trigger"
 	v2Client "github.com/fnproject/fn_go/clientv2"
 	clientApps "github.com/fnproject/fn_go/clientv2/apps"
-	clientFns "github.com/fnproject/fn_go/clientv2/fns"
-	clientTrigs "github.com/fnproject/fn_go/clientv2/triggers"
 	models "github.com/fnproject/fn_go/modelsv2"
 	"github.com/urfave/cli"
 )
@@ -150,7 +148,7 @@ func (p *deploycmd) deploy(c *cli.Context) error {
 	}
 
 	app, err := apps.GetAppByName(p.clientV2, appName)
-	if _, ok := err.(*clientApps.GetAppNotFound); ok && p.createApp {
+	if _, ok := err.(apps.NameNotFoundError); ok && p.createApp {
 		app = &models.App{
 			Name: appName,
 		}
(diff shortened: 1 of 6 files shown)


rdallman:
not having a good time trying to run tests locally... will fix CI up tonight/tomorrow if it's still broken (WIP then..)

commands/deploy.go
```diff
@@ -187,34 +220,21 @@ func (p *deploycmd) deploySingle(c *cli.Context, appName string, appf *common.Ap
...
 		}
-		if appf != nil {
-			if dir == wd {
-				setFuncInfoV20180708(ff, appf.Name)
-			}
+		if isAppYAML && dir == wd {
+			// TODO(reed): why are we setting this any differently than anything else? this is magical.
+			// for app.yaml that don't provide a func.yaml with a name, we use '$PWD-root' instead of '$PWD'
+			// for the name of the function - this is really weird. remove?
+			setFuncInfoV20180708(ff, app.Name)
```
rdallman: i find this behavior really weird, see https://github.com/fnproject/cli/pull/533/files#diff-1b0a895253821e350d6b12ecde4d6975L318 - do you think we can / should remove it and just name $PWD like everything else that doesn't provide a name in the func.yaml? this really tripped me up.

rdallman: this comment is inaccurate, it's actually `$appname-root` instead of `$PWD` that's happening - double trippy

riconnon: I agree, but it needn't be part of this PR.

rdallman: i'm cool with that, thanks

rdallman:
thanks @riconnon

PR #537, opened 2019-02-18
Python function build improvements

denismakogon:
This PR contains two things:

1. `setup.py` - this change allows functions to be packaged as redistributable python libraries. So, there are 2 things: `requirements.txt` as a placeholder for all deps references and `setup.py` that sets up the whole project. With first Fn CLI installs all dependencies into python path, with the second one, Fn CLI will add function's packages/modules into python path as well.
No impact on current behaviour, backwards-compatible change.

2. `.pip_cache` - with this change, Fn CLI will allow developers to use cached dependencies during function's build, there are reasons for this: some python libs may be stored within corp. networks, some of the dependencies being developed locally and not yet published anywhere.
No impact on current behaviour, backwards-compatible change.

### DIFF
--- a/langs/python.go
+++ b/langs/python.go
@@ -71,8 +71,8 @@ func (h *PythonLangHelper) DockerfileBuildCmds() []string {
 	if exists("requirements.txt") {
 		pip_cmd := `RUN pip3 install --target /python/  --no-cache --no-cache-dir`
 		if exists(".pip_cache") {
-			r = append(r, "ADD .pip_cache /function/")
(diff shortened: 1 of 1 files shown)


denismakogon:
Let’s wait on this, need couple tests to run.

denismakogon:
How to test this install from cache:
- `pip download -r requirements.txt -d .pip_cache`
- `fn --verbose build`

denismakogon:
@rdallman can you please check again?

rdallman:
LGTM

denismakogon:
thanks

PR #538, opened 2019-02-19
clean up invoke output

rdallman:
* no more async output catch (not a thing, at the moment)
* we can get fn errors that have a call id, so attempt to unwind them first
instead of just outputting them in json format. this should still output
user's function output even for non-200 responses the same, unless they use
the same json we do for errors which are non-empty, in which case adding some
window dressing from the cli doesn't seem horrible?
* adds a \n to the output of the cli invoke to make terminals more happy with it

below displays the issues we had previously where we were showing the full json
error from fn instead of unraveling it to show to the user, also we were
previously showing the help message if the function invoke returned >400,
which could denote a successful invocation that simply returns a >400 (from a
user perspective) - cleaned both of these up, and merged the errors into one
line instead of 2. I'm open to feedback on formatting the error, I'm tempted
to just do `Error invoking function: 504 Timed out` - open to other ideas.

master:

```
✗: cli invoke testapp hello
{"message":"Timed out"}

Fn: Error calling function: status 504

See 'fn <command> --help' for more information. Client version: 0.5.51
```

now:

```
✗: cli invoke testapp hello
Error invoking function. status: 504 message: Timed out
```

success output is the same:

```
 ✗: cli invoke testapp hello
{"message":"Hello World"}
```

except now if the function doesn't output a \n last, then the cli will, this
closes #1408 so that we play nice with the terminal

### DIFF
--- a/client/invoke.go
+++ b/client/invoke.go
@@ -16,7 +16,7 @@ import (
 )
 
 const (
-	FN_CALL_ID             = "Fn-Call-Id"
+	CallIDHeader           = "Fn-Call-Id"
 	MaximumRequestBodySize = 5 * 1024 * 1024 // bytes
 )
 
@@ -37,12 +37,7 @@ type apiErr struct {
 	Message string `json:"message"`
 }
 
-type callID struct {
-	CallID string `json:"call_id"`
-	Error  apiErr `json:"error"`
-}
-
-func Invoke(provider provider.Provider, invokeUrl string, content io.Reader, output io.Writer, method string, env []string, contentType string, includeCallID bool) error {
+func Invoke(provider provider.Provider, invokeURL string, content io.Reader, output io.Writer, method string, env []string, contentType string, includeCallID bool) error {
 
 	method = "POST"
 
@@ -52,13 +47,13 @@ func Invoke(provider provider.Provider, invokeUrl string, content io.Reader, out
 	if content != nil {
 		b, err := ioutil.ReadAll(io.LimitReader(content, MaximumRequestBodySize))
 		buffer := bytes.NewBuffer(b)
-		req, err = http.NewRequest(method, invokeUrl, buffer)
+		req, err = http.NewRequest(method, invokeURL, buffer)
 		if err != nil {
 			return fmt.Errorf("Error creating request to service: %s", err)
 		}
 	} else {
 		var err error
-		req, err = http.NewRequest(method, invokeUrl, nil)
+		req, err = http.NewRequest(method, invokeURL, nil)
 		if err != nil {
 			return fmt.Errorf("Error creating request to service: %s", err)
 		}
@@ -86,10 +81,10 @@ func Invoke(provider provider.Provider, invokeUrl string, content io.Reader, out
 	}
 
 	resp, err := httpClient.Do(req)
-
 	if err != nil {
-		return fmt.Errorf("Error invoking fn: %s", err)
+		return fmt.Errorf("Error invoking function: %s", err)
 	}
+	defer resp.Body.Close()
 
 	if logger.DebugEnabled() {
 		b, err := httputil.DumpResponse(resp, true)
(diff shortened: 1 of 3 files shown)


commands/invoke.go
```diff
@@ -116,5 +115,12 @@ func (cl *invokeCmd) Invoke(c *cli.Context) error {
 		}
 	}
 
-	return client.Invoke(cl.provider, invokeURL, content, os.Stdout, c.String("method"), c.StringSlice("e"), contentType, c.Bool("display-call-id"))
+	err := client.Invoke(cl.provider, invokeURL, content, os.Stdout, c.String("method"), c.StringSlice("e"), contentType, c.Bool("display-call-id"))
+	// we don't want to show the help message if invoke fails, just copy error to
+	// stderr but don't return the error from Invoke. also exit early here
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err.Error())
+		os.Exit(1)
```
rdallman: are there any thoughts about this vs getting rid of https://github.com/fnproject/cli/blob/master/main.go#L246-L249 ?

for invoke specifically I think it's really weird for us to garble up the output of users functions that don't really error in the context of fn (the function runs) with client version and all that, but I'm not opposed to getting rid of all that output altogether (my cli philosophy usually doesn't align with the corporate ones, however, so leaving this open...)

rdallman:
closes https://github.com/fnproject/fn/issues/1408

rdallman:
closes https://github.com/fnproject/fn/issues/1408

peterj:
Is there an option for JSON output in case of an error? I am thinking about scenarios where I invoke the function and want to check/parse the error message.

client/invoke.go
```diff
@@ -99,35 +94,52 @@ func Invoke(provider provider.Provider, invokeUrl string, content io.Reader, out
...
+			// this is likely from fn, so unravel this...
+			return fmt.Errorf("Error invoking function. status: %v message: %v", resp.StatusCode, msg.Message)
+		}
+
+		// read anything written to buffer first, then copy out rest of body
+		body = io.MultiReader(&b, resp.Body)
 	}
 
+	// at this point, it's not an fn error, so output function output as is
+	// TODO we should give users the option to see a status code too, like call id?
```
peterj: I'd vote yes, assuming the additional information can help them to investigate/figure out what the error is.

commands/invoke.go
peterj: is the behavior here that if the invoke fails, we should the Fn:, Client Version etc? If so, I don't think we should be showing that.

rdallman:
> Is there an option for JSON output in case of an error? I am thinking about scenarios where I invoke the function and want to check/parse the error message.

we have a json output flag for some cli commands at least, this might align with that. could check if that flag is set and give a json. what i'm thinking is that we might have to add a json format for this endpoint altogether, eg `{"headers":{"x":["y"]}, "status":200,"body":"yo"}` - this seems useful, it's hard to get headers out of invoke right now, which i've found painful personally. thanks for this suggestion. I'll add this to this PR.

client/invoke.go
rdallman: think https://github.com/fnproject/cli/pull/538#issuecomment-468371086 is a good spot for this, we can keep the default plaintext to just the body for now at least

peterj: sounds good. I just remember I was the one who added the json flag a while ago :) Might have to go through the commands again to see if there's anything I missed

commands/invoke.go
rdallman: this has improved with https://github.com/fnproject/cli/pull/538/commits/ca67a7b284d3de29190c69cbd39445fab667a792 and this block has gone away. now we only error and show the crap if there's issues building the request or talking to the service [without getting a response], and it's more clear what we do with output regardless of error status

rdallman:
https://github.com/fnproject/cli/pull/538/commits/ca67a7b284d3de29190c69cbd39445fab667a792 improves a few things, addresses comments. for the 'i want json error' case, now there's a json output format for the endpoint, in which the entire raw body (json) will be (parse json twice, it's meh, but it's ok), users can then do as they please with it.

rdallman:
thanks, gonna roll with this... shouldn't break anyone unless they were very exact about parsing errors out

denismakogon:
sorry, a bit too late, as we're changing the behaviour of the ` invoke ` cmd, can we also take care of HTTP headers usage? The use of that is not intuitive at all.

rdallman:
@denismakogon can you be more specific? you mean for adding headers to the request or getting them out?

denismakogon:
As part of the `fn invoke` I'd like to do 2 things:
1. As a user I'd like to specify HTTP headers like `fn invoke blah blah --http-header 'Blah: blah'

Optional (but good to use with OCI cURL)
2. As a user, I'd like to get annotation URL like `fn invoke blah blah --invoke-annotation-only` (don't blame me, haven't thought about better option) or anything that might be better to get an invoke annotation, so, I can wire that to OCI cURL.

rdallman:
thanks, that seems unrelated to this PR, but I agree, esp about headers. could you file an issue or 2 for those, please?

PR #541, opened 2019-02-21
small fixes to deploy

rdallman:
* if a user deploys a single function in a directory with an app.yaml and a
func.yaml that doesn't supply a name, this will now be named `$PWD` rather
than `$appname-root`, just like any other function that doesn't supply a name
in its func.yaml
* if a user deploys all with a function in the root directory that does not
supply a name in its func.yaml, this will now be named `$PWD` rather than
`$appname-root`, just like any other function that doesn't supply a name in
its func.yaml
* deploy all was using `--dir` which was an undocumented flag to specify the
`--working-dir` (as far as I can tell), this now aligns with deploy's help and
deploy single to use `--working-dir` for deploy all
* doing one less file read of the func yaml on deploy single now to check
version, woo... moved version check into parse step for func files.

I believe the reason for the old behavior was for apps to be able to have a
root route `/` and have that as the default behavior when specifying no
function name in the app root directory - now that names and routes aren't
tangled into the same thing, and we weren't using that policy anymore anyway,
I think it makes sense to move to using the $PWD name when not specifying a
name in the func.yaml like we are everywhere else, for consistency if
anything.

### DIFF
--- a/commands/deploy.go
+++ b/commands/deploy.go
@@ -224,7 +224,12 @@ func (p *deploycmd) deployAll(c *cli.Context, app *models.App) error {
 	if c.String("working-dir") != "" {
 		dir = c.String("working-dir")
 	} else {
-		dir = wd
+		// if we're in the context of an app, first arg is path to the function
+		path := c.Args().First()
+		if path != "" {
+			fmt.Printf("Deploying function at: /%s\n", path)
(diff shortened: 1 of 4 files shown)


rdallman:
thanks!

PR #542, opened 2019-02-27
this removes func format from everything

rdallman:
format is dead now, so we can remove all the flags and docs about it here.

technically, we should preserve this in old functions format to migrate,
however since there's no field to migrate to, it doesn't seem worth the
cognitive baggage?

there's a test that all our runtimes work off our generated func.yaml and
boilerplate, so that should prove that this change is ok.


rdallman:
thank you

PR #543, opened 2019-02-27
add syslog_url from app.yaml

rdallman:
right now in the cli the only way to set syslog url is via update or create
with the --syslog-url flag. this adds support in deploy when syslog_url is
present in app.yaml

### DIFF
--- a/common/appfile.go
+++ b/common/appfile.go
@@ -23,10 +23,10 @@ var (
 
 // AppFile defines the internal structure of a app.yaml/json/yml
 type AppFile struct {
-	Name string `yaml:"name,omitempty" json:"name,omitempty"`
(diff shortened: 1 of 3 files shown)


common/appfile.go
```diff
@@ -23,10 +23,10 @@ var (
 
 // AppFile defines the internal structure of a app.yaml/json/yml
 type AppFile struct {
```
denismakogon: that's happening again, can we reuse Fn models in the CLI because we had to manually copy-paste from server code?

rdallman: the fn models don't generate the yaml tags at the moment :/

rdallman:
thanks @denismakogon

PR #544, opened 2019-03-01
Add stricter func name validation

### LINKED ISSUE #503: CLI should be more strict about function names
`fn init`doesn't test if a function name has space(s) in its name or if the name starts/end with a separator. Separators should also be checked.

Ex. the following functions will be successfully bootstrapped by `init` but will fail at the `build` stage given their names doesn't comply with Docker naming convention.

```
fn init --runtime go "  "
fn init --runtime go "john doe"
fn init --runtime go "...."
fn init --runtime go .separator
fn init --runtime go separator.
```
and with other separators : `-`, `_`,`--`, `__`

peterj:
The valid function names contain a combination of lowercase letters, numbers
and separators (., -, _). Name can only start and end with a letter or
a number and there can only be one separator in the row (e.g. my---funcname
is not valid, however my-func_name is valid).

Fixes #503


init_test.go
```diff
@@ -58,17 +58,90 @@ func TestInit(t *testing.T) {
...
+		{
+			name:        "multiple different separators in row",
+			value:       "my_.funcblah",
+			errExpected: true,
+		},
+		{
+			name:        "multiple separators in the name",
+			value:       "my_func-blah.test",
+			errExpected: false,
+		},
```
denismakogon: there are more cases that suppose to be covered like `.{name}` that evaluates to the following image tag`{registry}/.{name}:{version}` <--- that's not a valid case.

So, i guess, names that start with `*`, `.`, `-` are not valid as well.

peterj: I have the `.myfunc` test case. I can add the other cases as well.

peterj: And I am pretty sure there are other cases as well (&, ~@#$%^&*...), but I don't think we actually need to test for all of them :)

peterj: Added those tests.

commands/init.go
```diff
@@ -397,14 +398,14 @@ func (a *initFnCmd) bindFn(fn *modelsV2.Fn) {
...
+// lowercase characters, separated by ., _ or -.
 func ValidateFuncName(name string) error {
-	if strings.Contains(name, ":") {
-		return errors.New("Function name cannot contain a colon")
-	}
-	if strings.ToLower(name) != name {
-		return errors.New("Function name must be lowercase")
+	// Match multiple groups of lowercase charaters and numbers, followed by 0 or 1 instances
+	// of the separators; the string also needs to end with a character or a number (not separator)
+	isValid := regexp.MustCompile(`^([a-z0-9]+[._-]?)*[a-z0-9]$`).MatchString
```
rdallman: we should compile this at init time and have a global var to use just to call MatchString, regex compile is slow and this func may have repeated invocations (eg in `deploy --all`)

peterj:
Thanks for the explanation! I figured there would be multiple places for this :) 

I am thinking perhaps fixing the b) as part of this PR as well and have a separate issue for a) and unifying it. 

How does that sound @rdallman ?

rdallman:
sounds good, thanks. I should've split a) into two parts, sorry - I think here we would be well served to validate the image name (ie after constructing from name) across init, build, bump, push and deploy

Issue #549, opened 2019-03-05
automate homebrew releases


carimura:
we're 30 days behind right now. not good.

carimura:
also `brew update` might be needed in docs

denismakogon:
in docs, tutorials and CLI README as well.
cc @michael-w-williams

devcamcar:
https://docs.brew.sh/How-To-Open-a-Homebrew-Pull-Request

denismakogon:
@carimura done

carimura:
and we're back. turns out there is no automation still.. see here for how we could do this:

https://github.com/yarnpkg/yarn/blob/a67d02f2e3d2996fda439f1ba0f0d25e4fa33d9d/scripts/update-homebrew.sh#L38-L45

https://github.com/Linuxbrew/brew/blob/master/docs/How-To-Open-a-Homebrew-Pull-Request.md

Probably do this in Circle. Or GitHub actions. :)

Daniel15:
I wrote the `update-homebrew.sh` script that Yarn uses to push release to Homebrew (linked above)... Let me know if you have any questions about it :)

carimura:
@Daniel15 fancy! thanks will do.

PR #550, opened 2019-03-05
Extend homebrew command

denismakogon:
The CLI has a pretty high release flow (each commit triggers a release), so, we need to recommend people to update homebrew repository info.

Closes: #549

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 ## Install
 
 ```sh
-brew install fn
+brew update && brew install fn
 ```
 or
 


carimura:
LGTM but not sure it should close #549 (if you want added to docs also)

denismakogon:
@carimura let me explain it, initially, we thought (me and @devcamcar) that homebrew Fn releases were outdated since 0.5.45, but it appeared that homebrew FN releases are fine just one minor version behind.
So, we can't blame homebrew for being late, because it's almost the same for every piece of software that is stored in public repositories if your package management tool has an old (stored locally) package references - please update it <--- that's what happened to @devcamcar.

In terms of the release schedule, one minor version delay is not critical, that's why I think this PR closes #549 (Fn CLI releases already automated but slightly outdated). As far as I understand, we have zero control over the pace of the homebrew package index update.

denismakogon:
let's open separate tickets within the corresponding repos (docs, tutorials).

denismakogon:
@carimura leaving big green "merge" button for you. :click-click-click:

devcamcar:
Also, for the record, they did just update the Fn homebrew formula 2 days ago, but it should be automated.  It's only being updated at random intervals, usually 1-3 months.

https://github.com/Homebrew/homebrew-core/commits/master/Formula/fn.rb

carimura:
@denismakogon can you please make sure a new ticket (or PR) is created for updating docs before closing #549? That's my point. I don't want to drop the hand off when follow ups are needed. thanks.

And ya there's a pretty big gap in versions on brew causing big delays.

PR #552, opened 2019-03-08
Revert "clean up invoke output (#538)"

harryjsmith:
This reverts commit ce8ab8a531678b4c62aa7ab86df9a564c0e4eb41. The changes in the commit caused invokes with a payload to fail with 401 Not Authenticated. Initial investigation seems to show the body might not be getting sent correctly. Without the offending commit, `echo -n '{name: foo}' | DEBUG=1 fn invoke harry-test hello` shows `{name: foo}` as the body of the request in the debug output. With the changes in the commit, the same CLI command shows
```
b
{name: foo}
0

``` 
as the body in the debug output. Reverting the commit for now until the issue is fixed.

### DIFF
--- a/client/invoke.go
+++ b/client/invoke.go
@@ -1,8 +1,11 @@
 package client
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"net/http/httputil"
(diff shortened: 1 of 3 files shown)


rdallman:
sorry, will try to add a test for this one, thanks for debugging.

PR #562, opened 2019-04-01
Fix: adding ability to update function's image

denismakogon:
Before this commit, the CLI provided almost all Fn model attributes except function's image. Meaning that users weren't able to update function with another image using:
```
    $ fn update fn
```

Before:
```
fn update fn -h

MANAGEMENT COMMAND
  fn update function - Update a function in application
    
USAGE
  fn [global options] update function [command options] <app-name> <function-name> 
    
DESCRIPTION
  This command updates a function in an application.
    
COMMAND OPTIONS
  --memory value, -m value  Memory in MiB (default: 0)
  --config value, -c value  Function configuration
  --timeout value           Function timeout (eg. 30) (default: 0)
  --idle-timeout value      Function idle timeout (eg. 30) (default: 0)
  --annotation value        Function annotation (can be specified multiple times)
```

Now:
```
fn update fn -h

MANAGEMENT COMMAND
  fn update function - Update a function in application

USAGE
  fn [global options] update function [command options] <app-name> <function-name>

DESCRIPTION
  This command updates a function in an application.

COMMAND OPTIONS
  --memory value, -m value  Memory in MiB (default: 0)
  --config value, -c value  Function configuration
  --timeout value           Function timeout (eg. 30) (default: 0)
  --idle-timeout value      Function idle timeout (eg. 30) (default: 0)
  --annotation value        Function annotation (can be specified multiple times)
  --image value             Function image

```

Fix by itself is very trivial, because all necessary code is already there, the only thing that was missing is the CLI option.

### DIFF
--- a/objects/fn/fns.go
+++ b/objects/fn/fns.go
@@ -49,6 +49,10 @@ var FnFlags = []cli.Flag{
 		Name:  "annotation",
 		Usage: "Function annotation (can be specified multiple times)",
 	},
+	cli.StringFlag{
+		Name:  "image",
+		Usage: "Function image",
+	},
 }
 var updateFnFlags = FnFlags
 


test/cli_crud_test.go
```diff
@@ -136,7 +136,8 @@ func TestFnUpdateValues(t *testing.T) {
 	}
 
 	invalidCases := [][]string{
-		{"--image", "fooimage:1.0.0"}, // image with no registry
+		// image with no registry is valid case for local development
+		// {"--image", "fooimage:1.0.0"}, // image with no registry
```
rdallman: @denismakogon any chance to add actual test here (above) https://github.com/fnproject/cli/pull/562/files#diff-8e679447ea0f37d6dcbf0abaaa9c5f36L119 ?

denismakogon: see #564

PR #567, opened 2019-04-04
no longer support logs and calls

rdallman:
this admittedly is not a great getting started xp, however, notably we've been
pointing users to syslog most of the time anyway in order to debug errors.
there is still the option to start fn in debug mode to get the logs, and we
can discuss that wherever, the last time it was brought up to discuss the only
voices that showed up were Jiminy Cricket.

see https://github.com/fnproject/fn/pull/1458 for details/discussion


objects/app/apps.go
```diff
@@ -78,9 +78,6 @@ func getApps(c *cli.Context, client *fnclient.Fn) ([]*modelsv2.App, error) {
 		resApps = append(resApps, resp.Payload.Items...)
 
 		n := c.Int64("n")
-		if n < 0 {
-			return nil, errors.New("Number of calls: negative value not allowed")
-		}
```
rdallman: there's 3 changes that all look like this. it looks like the error was copied from calls and i thought it was confusing, but I don't understand why it's here to begin with so I've just removed it. providing n < 0 will behave exactly the same as the default of n=0 - just getting one page -  so I see no reason to error here just because little Bobby Tables thinks it'd be fun to see if we do validation of negative numbers, we can just return a page and peace... stick that in your pipe and smoke it, Bobby

skinowski: unlikely case for large n < 0, howManyMore can overflow back to a positive value. But it's fine to remove it.

PR #569, opened 2019-04-08
lang runtime help string: sort and unique

kppk:
This changes the way the `fn init --help` list runtime languages.

**before:**

`--runtime value                Choose an existing runtime - go, java8, java11, java, node, python, python3.6, python, python3.7.1, ruby, kotlin`

_unsorted, python twice_ ^^

**after:**

`--runtime value                Choose an existing runtime - go, java, java11, java8, kotlin, node, python, python3.6, python3.7.1, ruby`

_sorted, unique_ ^^


commands/init.go
```diff
@@ -116,7 +117,16 @@ func langsList() string {
...
-	return strings.Join(allLangs, ", ")
+	sort.Strings(allLangs)
+	// remove duplicates
+	var allUnique []string
+	for i, s := range allLangs {
+		if i > 0 && s == allLangs[i-1] {
+			continue
+		}
+		allUnique = append(allUnique, s)
+	}
```
rdallman: hmm, this seems like a bug. i'm good with sorting them, but in this case I'd expect a test to run that means we never end up with 2 langs that exist under the same name, at a minimum case being we need to add a version to the runtime name - it means we have dead code for a runtime, at least, would be better to get rid of it if we don't want it or to add a version. i'm guessing this code is here cuz it's happening? can we remove the offending lang or affix a version to it?

kppk: oh yeah, good point, will fix it 👍

kppk: Interesting, it is hardcoded for java:
https://github.com/fnproject/cli/blob/62c68e49e84c9934730d2fc208adabbace9472f9/langs/java.go#L37

whereas in case of python if we use 'python' as runtime it will pick up the first python helper registered:
https://github.com/fnproject/cli/blob/62c68e49e84c9934730d2fc208adabbace9472f9/langs/python.go#L50

so there should not be any dead code, both `python` and `python3.6` will choose python 3.6 helper, cause that one is registered first here https://github.com/fnproject/cli/blob/62c68e49e84c9934730d2fc208adabbace9472f9/langs/base.go#L17

I think we can make java similar and use the unique runtime names? maybe put a comment that the registration order matters?

rdallman: comment or test would be good to scream at us when we update if we goof maybe

Issue #333, opened 2018-07-17
As an application developer I would like to the option to recursively delete an application, 


rikgibson:
At the moment `fn delete app <app-name>` deletes an app and all of it's routes/functions/triggers.  This behaviour is a bit brutal and should instead be optional, and by default shouldn't be allowed. So instead, try to delete an app with any child elements by `fn delete app foo` would simply respond with `You cannot delete an application that contains functions or triggers.`

`fn delete app` can then be extended to support two additional options:

1) `--cascade` (or `--recursive`) that will delete all child elements in bottom-up order, with an initial `Application "foo" contains 3 functions and 5 triggers, are you sure you want to delete it?` message and confirmation.
 
2) `--impact` - this will output a list of elements that would be deleted if `fn delete app foo --cascade` were executed.

msgodf:
Perhaps the more conventional `--dry-run` might be more suitable than `--impact`?

hibooboo2:
Id also say use cascade not recursive...

rikgibson:
Updated above to reflect entirely-reasonable comments...

msgodf:
Is it difficult for the server to check whether an app has any associated functions or triggers? (i.e. whether it is or isn't empty).

gviedma:
I think we shouldn't do any cascading deletes server-side or API-side. Instead, we could have sugar in the CLI to retrieve all resources associated with a function, recursively deleting them via their APIs prior to deleting the app. This would also be the most portable across CLI providers, since not all implementations will be able to support cascading deletes server-side.

mantree:
@gviedma i think that is what's being proposed.
"The CLI would instead recursively call DELETE on each /trigger, /function and /app endpoint."

zootalures:
We don't need this now - super low priority!

PR #571, opened 2019-04-11
Feature: cascading deletion of resources Fixes issue #333

bencurrerburgess:
This addresses issue #333 
This PR adds two flags to delete commands for Applications and Functions objects:

- **recursive (r)**:  Will retrieve all sub-objects and attempt to delete them from the bottom up before deleting the primary resource (user is prompted to continue after all of the associated resources are listed)
- **force (f)**: Will silence the prompt from a recursive delete and just deletes all resources without further user input (this only has an effect when used in combination with the recursive flag)

### example cascading
command:
```
fn delete fn <app-name> <fn-name> -r
```
will result in this prompt:
```
You are about to delete the following resources:
   Functions:    1
   Triggers:     <number of associated triggers>
This operation cannot be undone
Do you wish to proceed? y/n: 
```
if the user replies 'y' then:
```
Trigger  <trigger-name>  deleted
Function  <fn-name>  deleted
```
if the user replies 'n' then nothing is deleted:
```
Cancelling delete
```

### example cascading & forced
command:
```
fn delete app -f -r <app-name> 
```
results in:
```
Trigger  <trigger-name>  deleted
Function  <fn-name>  deleted
App <app-nam> deleted
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,3 @@ data/
 .idea/
 **/*.iml
 **/.DS_Store
-/vendor
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ data/
 .idea/
 **/*.iml
 **/.DS_Store
+/vendor
--- a/common/common.go
+++ b/common/common.go
@@ -600,8 +600,8 @@ func GetFuncYamlVersion(oldFF map[string]interface{}) int {
 	return 1
 }
 
-// UserConfirmedMultiResourceDeleteion will prompt the user for confirmation to delete all the the resources
-func UserConfirmedMultiResourceDeleteion(aps []*modelsv2.App, fns []*modelsv2.Fn, trs []*modelsv2.Trigger) bool {
+// UserConfirmedMultiResourceDeletion will prompt the user for confirmation to delete all the the resources
+func UserConfirmedMultiResourceDeletion(aps []*modelsv2.App, fns []*modelsv2.Fn, trs []*modelsv2.Trigger) bool {
 
 	apsLen := len(aps)
 	fnsLen := len(fns)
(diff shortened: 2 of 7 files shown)


zootalures:
Would prefer -r over -c as it’s consistent with  (e.g rm)

zootalures:
LGTM but could you write a couple of e2e tests (e.g. in cli_crud_test.go) ?  thanks

bencurrerburgess:
> LGTM but could you write a couple of e2e tests (e.g. in cli_crud_test.go) ? thanks

Added this in now, though it would seem that the tests on master are broken already, but the test I have written passes. 👍

.gitignore
```diff
@@ -11,3 +11,4 @@ data/
 .idea/
 **/*.iml
 **/.DS_Store
+/vendor
```
hatchan: The git repository contains a checked in `vendor` directory. Are you sure this needs to be ignored?

common/common.go
```diff
@@ -595,3 +599,142 @@ func GetFuncYamlVersion(oldFF map[string]interface{}) int {
 	}
 	return 1
 }
+
+// UserConfirmedMultiResourceDeleteion will prompt the user for confirmation to delete all the the resources
+func UserConfirmedMultiResourceDeleteion(aps []*modelsv2.App, fns []*modelsv2.Fn, trs []*modelsv2.Trigger) bool {
```
hatchan: `Deleteion` should be spelled `Deletion`

common/common.go
```diff
@@ -595,3 +599,142 @@ func GetFuncYamlVersion(oldFF map[string]interface{}) int {
...
+		return false
+	} else {
+		fmt.Println("Unrecognised input, should be y/n")
+		fmt.Println("Cancelling delete")
+		return false
+	}
+	return true
+}
+
+// ListFnsAndTriggersInApp lists all the fucntions associated with an app and all the triggers associated with each of those functions
```
hatchan: `fucntions` should be `functions`

common/common.go
```diff
@@ -595,3 +599,142 @@ func GetFuncYamlVersion(oldFF map[string]interface{}) int {
...
+	fmt.Printf("Do you wish to proceed? Y/N: ")
+	reader := bufio.NewReader(os.Stdin)
+	input, _ := reader.ReadString('\n')
+	if strings.ToLower(input) == "y\n" {
+		return true
+	} else if strings.ToLower(input) == "n\n" {
+		fmt.Println("Cancelling delete")
+		return false
+	} else {
+		fmt.Println("Unrecognised input, should be y/n")
```
hatchan: The `y/n` should probably match the casing of the first query, so `Y/N`

common/common.go
```diff
@@ -595,3 +599,142 @@ func GetFuncYamlVersion(oldFF map[string]interface{}) int {
...
+		fmt.Println("   Functions:   ", fnsLen)
+	}
+	if trsLen > 0 {
+		fmt.Println("   Triggers:    ", trsLen)
+	}
+	fmt.Println("This operation cannot be undone")
+	fmt.Printf("Do you wish to proceed? Y/N: ")
+	reader := bufio.NewReader(os.Stdin)
+	input, _ := reader.ReadString('\n')
+	if strings.ToLower(input) == "y\n" {
```
hatchan: You might want to consider running this through `strings.TrimSpace` to get rid of any spaces and newlines (https://golang.org/pkg/strings/#TrimSpace)

PR #575, opened 2019-04-17
add --version flag to fn start

### LINKED ISSUE #574: RFE: add a flag to use a specific version of Fn Server
By default `fn start` will pull `fnproject/fnserver:latest` if it's not yet in the registry. It would be nice to add a flag to override the `latest` tag/version. This flag could be useful to use a specific version of _fnserver_, eg. debugging, demo, test, etc.

rdallman:
e.g. `fn start --version 0.3.690` to start a specific version of
`fnproject/fnserver` docker image

tested ^ and latest. I did not do a lot of validation, but we don't appear to
do much of the other args either (that's my excuse and i'm sticking to it!),
it will fail pretty obviously with 'image not found' if someone specifies the
wrong version and we can't too easily check that without registry calls (we
don't need that), though I concede we could check for semver, latest doesn't
fit that either and is most convenient to keep current behavior. alas.

closes #574


delabassee:
LGTM (and sorry for the long delay!)

PR #578, opened 2019-05-07
Remove dep support

rdallman:
we were maintaining support for the dep tool, but go mod has been in action for 18 months now, it's probably okay to drop support for this now. we also have and still have support for functions without any vendoring solution, I left a TODO that is a bit curious about this I guess, the use for it would be users that do not want to build with mod on but vendor their packages themselves, this use case is going away soon too as go mod becomes the default and we can change at that time. we also no longer need gopath if we exclusively use mod, which is enticing (inasmuch as it can be) - ie no more `/go/src/func` directory. 

small change to change our test func to use mod

small change to update extension builder helper / remove dep references


rdallman:
bump @denismakogon could you review this one too, please?

denismakogon:
can we document this change somehow? the upgrade path might not be clear.
I feel like this PR has to be very gentle with older go functions by showing an upgrade path.

rdallman:
hmm, could document it. where, though? can document here, for reference at least. 

after looking, we've only been writing go.mod file for 2 months (defaulting functions to mod). so, probably we should leave support for it for a bit longer. users that need to migrate their funcs from dep to mod will do `go mod init` which migrates dependencies over from dep to mod, then they can delete the Gopkg.* files. we could facilitate, I'm not keen on automatically migrating on behalf of users, however. 

i'll change this pr just to fix up the test func pretty much. originally, thought there was a bug here, but don't think so anymore.

rdallman:
ok @denismakogon i updated to only fix the builder to use mod and the test func to use mod, no compat issues now. thanks for the eyes

denismakogon:
> hmm, could document it. where, though?

short blogpost would be great, maybe @michael-w-williams can help with a short tutorial.

denismakogon:
should we add some sort of a screamer/warning message saying "please use go mode!" ?

michael-w-williams:
@denismakogon @rdallman If you blog about it, let me know and I can convert it into doc. If there isn't and obvious home for it, we will find one.

PR #580, opened 2019-05-07
add loud warning about prod

rdallman:
link to helm chart. our current operating docs kind of suck, would like to fix
this too, but alas. helm seems a better place to point anyway, the base issue
is just to inform people here


rdallman:
if you have a fever, I have a prescription...

PR #583, opened 2019-05-20
Bump Node, Ruby FDK versions in 'init' boilerplate

harryjsmith:
This bumps the version of the Ruby and Node FDKs to the latest, so that new functions generated with 'fn init' pull in the latest version of the FDK dependency

### DIFF
--- a/langs/node.go
+++ b/langs/node.go
@@ -1,7 +1,10 @@
 package langs
 
 import (
+	"encoding/json"
+	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"path/filepath"
 )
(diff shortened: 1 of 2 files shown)


crush-157:
@denismakogon We could make it '~> 0.0.20' which would mean it takes the latest version from 0.0.20 up to 0.1.0

harryjsmith:
> @denismakogon We could make it '~> 0.0.20' which would mean it takes the latest version from 0.0.20 up to 0.1.0

I think from a user perspective it's nice to have the dependency fixed at a particular version once the Gemfile has been generated, especially given we don't use semver properly across the FDKs. I've pushed a changed so that the Ruby boilerplate will now behave the same as the Java one - it makes a call out to rubygems.org to get the latest FDK version then locks the Gemfile at that version. I'll push a similar change for Node.

riconnon:
This looks good from a point of view of bringing the others up to meet the experience with the Java FDK, but ultimately it would be nice to see the init images feature used for this stuff.
FDKs can push init images which output boilerplate aligned with the init image version and the CLI can just use `:latest` of the init image unless told otherwise.

riconnon:
Does this leave a gap in the handling of Go and Python?
These two differ in that they do not specify the version at all.

harryjsmith:
@riconnon Agreed on both points. For now, I'm going to merge this and we can do those in separate PRs.

PR #591, opened 2019-08-12
Support Resource Copy Into Fn Docker Image via func.yaml

bepearso:
Allows support for copying resources into the container via the `func.yaml` file. 
Eg.
```
copy_resources:
  - resource: 
      from: build-resource/wallet
      to: /wallet
  - resource: 
      from: test.sh
      to: /
```


zootalures:
Any chance you could look at writing an end-to-end test for this: 

See https://github.com/fnproject/cli/tree/master/test for examples

common/funcfile.go
```diff
@@ -51,6 +51,15 @@ type Expects struct {
 	Config []inputVar `yaml:"config" json:"config"`
 }
 
+type paths struct {
+	From string `yaml:"from,omitempty" json:"from,omitempty"`
+	To string `yaml:"to,omitempty" json:"to,omitempty"`
+}
+
+type CopyResources struct {
```
gviedma: Since this struct is not used outside of this package, it can be made private by renaming to `copyResources`.

common/common.go
```diff
@@ -380,6 +380,16 @@ func writeTmpDockerfile(helper langs.LangHelper, dir string, ff *FuncFile) (stri
 		dfLines = append(dfLines, fmt.Sprintf("FROM %s", ri))
 		dfLines = append(dfLines, "WORKDIR /function")
 		dfLines = append(dfLines, helper.DockerfileCopyCmds()...)
+		
+		resources := ff.Copy_resources
+
+		for _, resource := range resources {
+			if Exists(resource.Resource.From){
```
gviedma: When checking if the file exists, you need to differentiate between absolute and relative paths. For absolute paths (those with a leading `/`), the above will work. For relative paths, you will need to resolve the full path prior to calling `Exists`. See for example, https://github.com/fnproject/cli/blob/17c045e83ccfc0ab8c39902c67f9858de8097ffb/common/common.go#L154

common/funcfile.go
```diff
@@ -101,6 +113,8 @@ type FuncFileV20180708 struct {
 	Timeout      *int32 `yaml:"timeout,omitempty" json:"timeout,omitempty"`
 	IDLE_timeout *int32 `yaml:"idle_timeout,omitempty" json:"idle_timeout,omitempty"`
 
+	Copy_resources []CopyResources `yaml:"copy_resources,omitempty" json:"copy_resources,omitempty"`
```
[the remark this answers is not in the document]
zootalures: Why have two layers here (i.e CopyResources.Resource) vs ? 

Could this be: 

```
copy_resources: 
     - from: myfile
       to: dest
     - from: otherfile
        to: dest2
````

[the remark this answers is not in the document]
Comment (gviedma) [reply common/funcfile.go]: I also think removing one layer of nesting here could improve readability.

File: test/cli_copy_resource_test.go [hunk 4]
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"strings"
+	"testing"
+
+	"github.com/fnproject/cli/testharness"
+)
+
+func TestResourceCopy(t *testing.T) {
+
+	const testFuncSource = `package main
```
gviedma: Can we move the function and yaml sources to their own directory under `test` rather than including them inline? See for example https://github.com/fnproject/cli/tree/206ff749b2f2251e90e89f37bba5d4739d82ed56/test/simplefunc

test/cli_copy_resource_test.go
```diff
@@ -0,0 +1,65 @@
...
+	h.WithFile("testfile.txt", testFile, 0644)
+	h.WithFile("go.mod", gomod, 0644)
+
+	h.Fn("deploy", "--app", appName, "--local", "--create-app").AssertSuccess()
+
+	res := h.Fn("invoke", appName, "filefn")
+
+	res.AssertSuccess()
+	if strings.TrimSpace(res.Stdout) != testFile {
+		t.Fatalf("Exepected result to be '%s', got '%s'", testFile, res.Stdout)
```
gviedma: Typo at Expected

common/common.go
```diff
@@ -380,6 +380,15 @@ func writeTmpDockerfile(helper langs.LangHelper, dir string, ff *FuncFile) (stri
 		dfLines = append(dfLines, fmt.Sprintf("FROM %s", ri))
 		dfLines = append(dfLines, "WORKDIR /function")
 		dfLines = append(dfLines, helper.DockerfileCopyCmds()...)
+		
+		resources := ff.CopyResources
+		for _, resource := range resources {
+			if Exists(resource.From){
+				dfLines = append(dfLines, fmt.Sprintf("COPY %s %s", resource.From, resource.To))
+			}else{
```
gviedma: Can you runt he linter? There are some minor problems with the formatting such as with the spacing.

common/common.go
```diff
@@ -434,6 +443,16 @@ func writeTmpDockerfileV20180708(helper langs.LangHelper, dir string, ff *FuncFi
 		dfLines = append(dfLines, fmt.Sprintf("FROM %s", ri))
 		dfLines = append(dfLines, "WORKDIR /function")
 		dfLines = append(dfLines, helper.DockerfileCopyCmds()...)
+		
+		resources := ff.CopyResources
+
+		for _, resource := range resources {
+			if Exists(resource.From) {
+				dfLines = append(dfLines, fmt.Sprintf("COPY %s %s", resource.From, resource.To))
```
zootalures: can you quote the destination with `strconv.Quote(resource.To)` to prevent command injection here.

shaunsmith:
I'm not convinced of the utility of this addition.  It only works for dynamically generated Dockerfiles which we've previously discussed removing in favour of using init-images.   Also, the currently generated Dockerfiles have inconsistent semantics, some copy all resources into the container image  (Node), and some don't (Java). 

I'd prefer we move in the direction of having the CLI invoke standard init-images in response to "--runtime" and remove the hard coded Go code that generates the dynamic Dockerfiles.

PR #592, opened 2019-08-22
Bugfix/cli loses context under concurrent operations

shreyagarge:
Fn cli loses context and uses default when concurrent cli operations are executed.
This is because every command loads the config, and the LoadConfiguration() method reads from the config file, sets context and _writes the context back_. This happens with every CLI operation, and there are no locks on these write operations. This causes concurrent CLI operations that happen to read while the file is being written to, pick up empty context, and the CLI then uses the default context (localhost), and it writes this empty context back after this operation, causing the context to get lost.

Fix : removed the writing config back bit from the loadconfig operation, leaving it only in the usectx action where it is actually needed. To prevent ever having an empty config file, the write operation is made atomic. 

note: The write operation is made atomic by using a tmp file and replacing it with the original config. This works perfectly on *nix systems as the rename operation is atomic on them, but not in windows. But it still is not an issue as we don't write config with every operation anymore.


config/config.go
```diff
@@ -171,10 +170,42 @@ func WriteConfigValueToConfigFile(key, value string) error {
...
-	err = WriteYamlFile(f.Name(), &configValues)
+	err = atomicwrite(f.Name(), &configValues)
 	if err != nil {
 		return err
 	}
 
 	return nil
 }
+
+func atomicwrite(file string, c *ContextMap) (err error) {
```
templecloud: So, the issue this is fixing, is that when using multiple processes (by invoking the Fn CLI from a shell as multplie times as background processes) this config file can attempt to be read by one process as another is writing to it. The config file  contains the information on what 'context' the CLI should be using, so in these situations comes back 'empty' which is the definition of the `default` context.

The solution presented here is to write to a tmp file first, then, move it. Subsequently make the file write an atomic operation.

Nice. Looks goo to me.

config/config.go
```diff
@@ -171,10 +170,42 @@ func WriteConfigValueToConfigFile(key, value string) error {
...
+	err = WriteYamlFile(f.Name(), c)
+	if err != nil {
+		return err
+	}
+
+	info, err := os.Stat(file)
+	if err != nil {
+		return err
+	} else {
+		_ = os.Chmod(f.Name(), info.Mode())
```
templecloud: Just a nit, but we ignore the error here. Is it not worth chucking this back too? 

NB :I don't know if it is possible , to set things up in a way that would cause this, but if anyone every want to use the CLI inside a container the sometimes things might not not set up quite right.

templecloud: Just has a look at this: it returns a 'PathError'. It is probably fine.

PR #593, opened 2019-08-23
Add fdk-go requirement to hello world mod file

ant1eicher:
I noticed a discrepancy between the requirements specified in the CLI boilerplate hello world `go.mod` and the `go.mod` specified in the Go FDK: https://github.com/fnproject/fdk-go/blob/master/examples/hello/go.mod. Since the example uses the FDK, I believe it should be specified as a requirement in the modules file.

### DIFF
--- a/langs/go.go
+++ b/langs/go.go
@@ -129,7 +129,8 @@ func myHandler(ctx context.Context, in io.Reader, out io.Writer) {
 }
 `
 
-	modBoilerplate = `
-module func
+	modBoilerplate = `module func
+
+require github.com/fnproject/fdk-go master
 `
 )


rdallman:
thanks. I'm okay with doing this though I'm a bit conflicted about this simply due to putting an actual version of the fdk into the go mod file (master, here). while the fdk-go does need to get real semantic versions, once it does, I'd assume that we would want to rely on `go mod` tool to actually deduce these upon first build - that is, `go build` will add the line proposed in this patch to the go mod file when build is run. this allows more flexibility over time where they are correctly locked into a real version of the fdk from when they created the function, whereas if master is specified this will get updated frequently and could (theoretically) break. the other case would be that users use dependencies that depend on the fdk and they may rely on a specific version of it which may need to be deduced and is less than master, this may require user intervention that could have been avoided - this isn't as important because we aren't adding mod files to go packages that already have them.

I guess the difference between the CLI boilerplate and the hello world in the examples is that hello world is a created example whereas the cli generated function is in the process of being created (go build has not yet been run), also the example in the fdk-go needs to specify the master version because we're using it in CI to make sure that the fdk hasn't broken anything. I think in the CLI over time we'll likely prefer a specific version to get cemented into user's go mod files (go mod does this for us) vs specifying master. in any case, since we don't have real version at this time, I can see how this seems wishy washy. anyway, just my $.02, perhaps others agree/disagree

ant1eicher:
@rdallman, I think you are right.

Assuming the workflow:

1. `fn init --runtime go gofn`, which generates the `go.mod` file
2. `fn deploy --app goapp`, which copies the source into the container and does a `go build` inside the container (thereby updating `go.mod` file *inside* the container only)

We could decide between the following options:

### latest
```
module func

require github.com/fnproject/fdk-go latest
```
Finds version `*1458ca84e01d`
```
$ fn --verbose deploy --app goapp --local
...
go: finding github.com/fnproject/fdk-go latest
go: downloading github.com/fnproject/fdk-go v0.0.0-20190716163646-1458ca84e01d
```

### master
```
module func

require github.com/fnproject/fdk-go master
```
does the same as *latest*:

```
$ fn --verbose deploy --app goapp --local
go: finding github.com/fnproject/fdk-go master
go: downloading github.com/fnproject/fdk-go v0.0.0-20190716163646-1458ca84e01d
```

### empty
```
module func
```
also does the same as *latest*:

```
$ fn --verbose deploy --app goapp --local
...
go: finding github.com/fnproject/fdk-go latest
go: downloading github.com/fnproject/fdk-go v0.0.0-20190716163646-1458ca84e01d
```

Given that all of these outputs are the same for the first build of the function, we should probably just discard this PR and leave the `go.mod` file empty of requirements. It would be nice if the Fn cli could detect an unspecified requirement and notify the user that it would be in their best interest to peg that version to prevent future breaking changes. I'll leave that for future consideration.

ant1eicher:
@rdallman I think @owencliffe's suggestion of using init images is a good one. I see that the Go FDK already builds an init image, but we don't publish it to Docker Hub, and we don't use it in the CLI's GoLanguageHelper. I can make branches to do those things, and we can discuss whether to go ahead.

PR #594, opened 2019-09-05
Decouple and tidy up init code

ant1eicher:
In preparation for using an init image as the default function generation image for the Go language helper, I moved some code around and wrote some tests.

Phase 2 will be to define an init image in the language helper, and call that instead of the boilerplate.

### DIFF
--- a/commands/init.go
+++ b/commands/init.go
@@ -20,13 +20,9 @@ it will print an error message then exit.
 */
 
 import (
-	"archive/tar"
-	"bytes"
 	"errors"
 	"fmt"
-	"io"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"sort"
 	"strings"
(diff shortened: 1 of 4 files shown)


commands/init.go
```diff
@@ -299,81 +257,34 @@ func (a *initFnCmd) init(c *cli.Context) error {
...
-
-	c1 := exec.Command("docker", "run", "--rm", "-e", "FN_FUNCTION_NAME="+a.ff.Name, initImage)
-	c1.Stderr = &c1ErrB
-	c1.Stdout = tarW
-
-	c1Err := c1.Start()
-	if c1Err != nil {
-		fmt.Println(c1ErrB.String())
-		return errors.New("Error running init-image")
+func (a *initFnCmd) doInitImage(err error, initImage string, c *cli.Context) error {
```
riconnon: Why do you take an error as a parameter then immediately overwrite it here? Since this is unexported and we don't need to match any specific function signature can we just remove the parameter?

ant1eicher: Good catch. I blame Goland refactoring :)

ant1eicher:
Manual testing:

```
$ ./fn init --init-image fnproject/fdk-go-init tmp
Creating function at: ./tmp
Running init-image: fnproject/fdk-go-init
func.yaml created.

$ ls tmp
func.go   func.yaml go.mod

$ cat tmp/go.mod
module tmp

go 1.12

require github.com/fnproject/fdk-go v0.0.0-20190904171902-ab311c7b1289
```

PR #596, opened 2019-09-26
Modify init-image to pass through proxy settings and use StdoutPipe to prevent failing init image execution from blocking forever

ant1eicher:
Pass through http proxy variables to init-image docker invocation. Use StdoutPipe to prevent failing init image execution from blocking forever.

Testing:

**Error case (with bad proxy settings passed through):**

```
$ fn init --init-image fdk-go-init:latest tmp
Creating function at: ./tmp
Running init-image: fdk-go-init
Executing docker command: run --rm -e FN_FUNCTION_NAME=tmp -e https_proxy=http://foo.com:80 fdk-go-init:latest

Fn: Error running init-image: exit status 1. Stderr: 'go: creating new go.mod: module tmp
build tmp: cannot load github.com/fnproject/fdk-go: cannot find module providing package github.com/fnproject/fdk-go
'
```

**Success case (with no proxy settings passed through):**

```
$ fn init --init-image fdk-go-init:latest tmp
Creating function at: ./tmp
Running init-image: fdk-go-init:latest
Executing docker command: run --rm -e FN_FUNCTION_NAME=tmp fdk-go-init:latest
func.yaml created.

$ ls tmp
func.go   func.yaml go.mod

$ cd tmp

$ fn --registry test deploy --local --no-bump --app goapp
Deploying tmp to app: goapp
Building image test/tmp:0.0.1 ..........
Updating function tmp using image test/tmp:0.0.1...
Successfully created function: tmp with test/tmp:0.0.1

$  fn invoke goapp tmp
{"message":"Hello World"}
```

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -751,36 +751,50 @@ func ListTriggersInFunc(c *cli.Context, client *fnclient.Fn, fn *modelsv2.Fn) ([
 func DockerRunInitImage(initImage string, fName string) error {
 	fmt.Println("Running init-image: " + initImage)
 
-	// Run the initImage
-	var c1ErrB bytes.Buffer
-	tarR, tarW := io.Pipe()
+	args := []string{"run", "--rm", "-e", "FN_FUNCTION_NAME=" + fName}
+	args = append(args, proxyArgs()...)
+	args = append(args, initImage)
 
-	c1 := exec.Command("docker", "run", "--rm", "-e",
-		"FN_FUNCTION_NAME="+fName,
-		"-e", "http_proxy="+os.Getenv("http_proxy"),
-		"-e", "https_proxy="+os.Getenv("https_proxy"),
-		"-e", "no_proxy="+os.Getenv("no_proxy"),
-		initImage)
-	c1.Stderr = &c1ErrB
-	c1.Stdout = tarW
-
-	c1Err := c1.Start()
-	if c1Err != nil {
-		fmt.Println(c1ErrB.String())
-		return errors.New("Error running init-image")
+	fmt.Printf("Executing docker command: %s\n", strings.Join(args, " "))
+	cmd := exec.Command("docker", args...)
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		return fmt.Errorf("Error attaching stdout pipe to init-image cmd: %v.", err)
 	}
 
-	err := untarStream(tarR)
+	err = cmd.Start()
 	if err != nil {
-		return errors.New("Error un-tarring the output of the init-image")
+		return fmt.Errorf("Error starting init-image: %v. Stderr: '%s'", err, stderr.String())
+	}
+
+	err = untarStream(stdout)
+	if err != nil {
+		return fmt.Errorf("Error un-tarring stdout from the init-image: %v", err)
+	}
+
+	err = cmd.Wait()
+	if err != nil {
+		return fmt.Errorf("Error running init-image: %v. Stderr: '%s'", err, stderr.String())
 	}
 
 	return nil
 }
 
+// Get any proxy environment variables
+func proxyArgs() []string {
+	args := make([]string, 0)
+	for _, a := range []string{"http_proxy", "https_proxy", "ftp_proxy", "no_proxy", "HTTP_PROXY", "HTTPS_PROXY", "FTP_PROXY", "NO_PROXY"} {
+		if val, ok := os.LookupEnv(a); ok {
+			args = append(args, "-e", a+"="+val)
+		}
+	}
+	return args
+}
+
 // Untars an io.Reader into the cwd
 func untarStream(r io.Reader) error {
-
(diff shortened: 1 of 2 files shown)


common/common.go
```diff
@@ -777,8 +782,22 @@ func DockerRunInitImage(initImage string, fName string) error {
...
+	res := make(chan bool)
 	for {
-		header, err := tr.Next()
+
+		var header *tar.Header
+		var err error
+
+		go func() {
+			header, err = tr.Next()
+			res <- true
```
skinowski: go-routine leak, if timer below fires then there's no consumer on this chan.

common/common.go
```diff
@@ -751,23 +751,28 @@ func ListTriggersInFunc(c *cli.Context, client *fnclient.Fn, fn *modelsv2.Fn) ([
...
 	tarR, tarW := io.Pipe()
 
-	c1 := exec.Command("docker", "run", "--rm", "-e", "FN_FUNCTION_NAME="+fName, initImage)
-	c1.Stderr = &c1ErrB
-	c1.Stdout = tarW
-
-	c1Err := c1.Start()
-	if c1Err != nil {
-		fmt.Println(c1ErrB.String())
+	cmd := exec.Command("docker", "run", "--rm", "-e",
```
skinowski: it's easier to use CommandContext() here: https://golang.org/pkg/os/exec/#CommandContext

common/common.go
```diff
@@ -751,23 +751,28 @@ func ListTriggersInFunc(c *cli.Context, client *fnclient.Fn, fn *modelsv2.Fn) ([
...
+		"FN_FUNCTION_NAME="+fName,
+		"-e", "http_proxy="+os.Getenv("http_proxy"),
```
skinowski: minor nit, it's probably better to not set these at all if env vars are not present. (as opposed to setting them to empty string)

common/common.go
ant1eicher: The original problem was that the `tr.Next()` blocks forever, and prevents the program from terminating.

Now, the timeout will fire, return an error, and the CLI will terminate the program (see above manual testing).

I'm attempting to follow this pattern: https://gobyexample.com/timeouts

Perhaps unsuccessfully? Could you suggest an alternative?

common/common.go
ant1eicher: Thanks for the pointer - CommandContext looks useful when you want to wrap a Command in a timeout. However, that's not the case here. In the error case, the command *does* return after ~1 minute, but it doesn't push any data onto stdout for the `untarStream` command to consume - thus resulting in `tr.Next()` blocking. Hence wrapping `tr.Next()` in a timeout as opposed to the Command.

Perhaps I'm not understanding your suggestion properly - I'll try to play around with this.

common/common.go
ant1eicher: Nvm. I think I've figured out a more elegant approach.

common/common.go
ant1eicher: Nvm. I think I've figured out a more elegant approach.

ant1eicher:
@skinowski thanks for the great feedback. I think I've come up with a better approach now. See updated PR.

skinowski:
There's slight change in buffering behavior here, where now we do store entire tar buf in memory. Previously this was being streamed possibly/likely reducing memory usage. Other than this concern, LGTM. I'll leave it to @riconnon @bencurrerburgess @JadeRedworth for a second look.

ant1eicher:
Thanks. Yes, the buffering change is intentional. The output from init image is quite small, and it allows us to verify a successful exit code before untarring the output.

ant1eicher:
> There's slight change in buffering behavior here, where now we do store entire tar buf in memory. Previously this was being streamed possibly/likely reducing memory usage. Other than this concern, LGTM. I'll leave it to @riconnon @bencurrerburgess @JadeRedworth for a second look.

I've been thinking about this a little more, and I think you're right to worry about the buffering @skinowski . Initially, I assumed that init image output would always be really small - a couple of KB at most, but then I realised that we *could* potentially make a large init image that contained some sort of framework, like rails or an app server. I'm working on a different solution which should be able to handle failed executions without blocking on waiting for stdout.

ant1eicher:
Hmm. Looks like there's a problem with the init image command on CircleCI. Unfortunately, I everything works fine locally, and I can't reproduce the issue. @skinowski or @riconnon could you please give me CircleCI write permissions, so that I can trigger a build with SSH access?

ant1eicher:
Moved over to using StdoutPipe pattern (https://golang.org/pkg/os/exec/#Cmd.StdoutPipe). Working on both Mac and Linux now.

common/common_test.go
```diff
@@ -26,3 +28,43 @@ func TestValidateImageName(t *testing.T) {
 		})
 	}
 }
+
+func Test_proxyArgs(t *testing.T) {
```
riconnon: Nit: There are a bunch of tools (eg. golint) which will complain about underscores in names. Might be nice to avoid.

riconnon: Scratch that... looks like it doesn't apply that rule to test functions.

PR #597, opened 2019-10-02
Revert "Support Resource Copy Into Fn Docker Image via func.yaml"

rikgibson:
Reverts fnproject/cli#591


shaunsmith:
Yes, let's revert this until the feature is more throughly thought through.

PR #601, opened 2019-10-28
Revendor in oracle provider support

jan-g:
Sign the opc-compartment-id header


riconnon:
@jan-g looks good, shall we also do a non-bugfix version bump on this?

PR #603, opened 2019-11-03
Rejig the Python copy commands

jan-g:
A small change to func.py can peturb a potentially larger layer
containing required libraries; shift the copy around to avoid this.

Add /function to the PYTHONPATH - this permits adding small sets of
other modules in a cheap & cheerful fashion alongside the func.py

The results here from a build are

```
/function/func.py & co.
/python/required-libraries
... the base image ...
```

as opposed to

```
/python/required-libraries
/function/func.py & co.
... the base image ...
```

### DIFF
--- a/langs/python.go
+++ b/langs/python.go
@@ -118,8 +118,8 @@ def handler(ctx, data: io.BytesIO=None):
 
 func (h *PythonLangHelper) DockerfileCopyCmds() []string {
 	return []string{
-		"COPY --from=build-stage /function /function",
 		"COPY --from=build-stage /python /python",
-		"ENV PYTHONPATH=/python",
+		"COPY --from=build-stage /function /function",
+		"ENV PYTHONPATH=/function:/python",
 	}
 }


langs/python.go
```diff
@@ -118,8 +118,8 @@ def handler(ctx, data: io.BytesIO=None):
 
 func (h *PythonLangHelper) DockerfileCopyCmds() []string {
 	return []string{
-		"COPY --from=build-stage /function /function",
 		"COPY --from=build-stage /python /python",
-		"ENV PYTHONPATH=/python",
+		"COPY --from=build-stage /function /function",
+		"ENV PYTHONPATH=/function:/python",
```
denismakogon: I feel like this is not the place to do the following change, FDK is where I believe this must be added to make sure that Python FDK CLI will be capable to do modules discovery starting from the root folder (that would be `/function`

jan-g: I'm not sure I follow this; the change works as-is. I can't find any use of `sys.path` ni the FDK itself - it looks to rely on this setting in order to operate.

denismakogon: I mean users must be aware that if they go with built-in dockerfile then the `PYTHONPATH` would contain not just the interpreter path but `/function` folder as well.

In order to make this process versatile and `PYTHONPATH`-independent, Python FDK must be capable to process the root folder of a function but in a manner of the delayed imports. Meaning, the root folder modules and packages must be initialized only when user's code gets executed or it will turn into a nightmare of imports profiling.

jan-g: This seems to behave okay under testing.

What I've seen from actual customer use is typically the addition of small helper modules - nothing large, but nonetheless inconvenient to pull into a single `func.py`.

Given that there's a need for this now - is it possible to act, then iterate?

ant1eicher:
Verified this works locally. I think it solves the current issues that users are having without introducing significant future issues.

PR #605, opened 2020-01-10
Fix atomicwrite for Windows

mattn:
os.Rename does not overwrite on Windows


mattn:
Can you take a look?

mattn:
Hi, currently, `fn use context` always fail since this issue. Windows can not rename/move file that handle of the file is opened.

PR #606, opened 2020-02-22
Add sample log message to the boilerplate code

srpurani:
Add sample log message lines to boilerplate code.

### DIFF
--- a/langs/go.go
+++ b/langs/go.go
@@ -105,6 +105,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log"
 
 	fdk "github.com/fnproject/fdk-go"
 )
@@ -125,6 +126,7 @@ func myHandler(ctx context.Context, in io.Reader, out io.Writer) {
 	}{
 		Msg: fmt.Sprintf("Hello %s", p.Name),
 	}
+	log.Print("Inside Go Hello World function")
 	json.NewEncoder(out).Encode(&msg)
 }
 `
--- a/langs/kotlin.go
+++ b/langs/kotlin.go
@@ -315,9 +315,12 @@ const (
 	helloKotlinSrcBoilerplate = `
 package com.fn.example
 
-fun hello(input: String) = when {
-    input.isEmpty() -> ("Hello, world!")
-        else -> ("Hello, ${input}")
+fun hello(input: String): String {
+    println("Inside Kotlin Hello World function")
+    return when {
+        input.isEmpty() -> ("Hello, world!")
+            else -> ("Hello, ${input}")
+    }
 }`
 
 	helloKotlinTestBoilerplate = `package com.fn.example
--- a/langs/python.go
+++ b/langs/python.go
@@ -95,6 +95,7 @@ func (h *PythonLangHelper) IsMultiStage() bool {
 const (
 	helloPythonSrcBoilerplate = `import io
 import json
+import logging
 
 from fdk import response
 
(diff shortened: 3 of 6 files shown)


langs/go.go
```diff
@@ -125,6 +126,7 @@ func myHandler(ctx context.Context, in io.Reader, out io.Writer) {
 	}{
 		Msg: fmt.Sprintf("Hello %s", p.Name),
 	}
+	log.Print("This is a sample log message.")
```
mrc: Just a nit: rather than logging `This is a sample log message.`, it could be `invoking function (Hello, world!)`, which (to me) looks more like auto-generated boilerplate and less like sample code.

sachin-pikle:
I agree with Matt .. In fact, I usually have an entering and existing ... something like the following .. 

```
        System.out.println("Entering Java HelloFunction::handleRequest()");

        String name = (input == null || input.isEmpty()) ? "world"  : input;
        
        System.out.println("Value of name is " + name);
        System.out.println("Exiting Java HelloFunction::handleRequest()");
```

sachin-pikle:
And for Python, without a flush=True we don't see any statements in the logs .. which trips most people!

```
    print("Entering Python Hello World handler", flush=True)
    .....
    print("Vale of name = ", name, flush=True)
    print("Exiting Python Hello World handler", flush=True)
 

```

srpurani:
> And for Python, without a flush=True we don't see any statements in the logs .. which trips most people!
> 
> ```
>     print("Entering Python Hello World handler", flush=True)
>     .....
>     print("Vale of name = ", name, flush=True)
>     print("Exiting Python Hello World handler", flush=True)
>  
> ```

Since we are using logging.getLogger(), it flushes the data.

langs/go.go
srpurani: Will make this change now.

srpurani:
> I agree with Matt .. In fact, I usually have an entering and existing ... something like the following ..
> 
> ```
>         System.out.println("Entering Java HelloFunction::handleRequest()");
> 
>         ....
>         
>         System.out.println("Value of name is " + name);
>         System.out.println("Exiting Java HelloFunction::handleRequest()");
> ```
> 
> Also, do we have a preference/recommendation on system.out versus system.err?

I tried with both System.out and System.err, there is no difference in the logs. We can keep this System.out.println().

I will change the log messages to "Invoking Java/Python/Ruby/... Hello World function"

langs/kotlin.go
```diff
@@ -317,7 +317,10 @@ package com.fn.example
 
 fun hello(input: String) = when {
     input.isEmpty() -> ("Hello, world!")
-        else -> ("Hello, ${input}")
+        else -> {
+          println("Inside Kotlin Hello World function")
```
sachin-pikle: @srpurani - For the Kotlin one, can you keep the print statement outside the if/else block?

langs/python.go
```diff
@@ -107,6 +108,7 @@ def handler(ctx, data: io.BytesIO=None):
     except (Exception, ValueError) as ex:
         print(str(ex))
```
sachin-pikle: @srpurani - while you are here, can you fix this one as well? Either add Flush or use logger?

langs/go.go
srpurani: Done.

langs/kotlin.go
srpurani: Done.

langs/python.go
srpurani: Done.

PR #607, opened 2020-03-09
Add Cloudshell support

roroco:
- Pickup version of fn go with support for cloudshell provider
- Read oci cli auth environment variable to determine what oracle provider to use


config/config.go
```diff
@@ -30,13 +31,37 @@ const (
...
 )
 
-var defaultRootConfigContents = &ContextMap{CurrentContext: "", CurrentCliVersion: ""}
-var DefaultContextConfigContents = &ContextMap{
-	ContextProvider:      DefaultProvider,
-	provider.CfgFnAPIURL: defaultLocalAPIURL,
-	EnvFnRegistry:        "",
+var defaultRootConfigContents = &ContextMap{CurrentContext: "default", CurrentCliVersion: Version}
+
+func DefaultContextConfigContents() (contextMap *ContextMap) {
```
mrc: Is this logic covered by any tests?

config/config.go
```diff
@@ -30,13 +31,37 @@ const (
...
+			ContextProvider: fn_go.OracleIPProvider,
+			EnvFnRegistry:   "",
+		}
+	} else {
+		contextMap = &ContextMap{
+			ContextProvider:      fn_go.DefaultProvider,
+			provider.CfgFnAPIURL: defaultLocalAPIURL,
+			EnvFnRegistry:        "",
+		}
+		viper.SetDefault(provider.CfgFnAPIURL, defaultLocalAPIURL)
```
mrc: How come `provider.CfgFNAPIURL` is only set for the default case (lines 58 and 61)? It's not clear why it's only being done for one (if someone added a new case, how do they decide whether this is necessary?) Maybe add a clarifying comment.

config/config.go
```diff
@@ -30,13 +31,37 @@ const (
...
-var DefaultContextConfigContents = &ContextMap{
-	ContextProvider:      DefaultProvider,
-	provider.CfgFnAPIURL: defaultLocalAPIURL,
-	EnvFnRegistry:        "",
+var defaultRootConfigContents = &ContextMap{CurrentContext: "default", CurrentCliVersion: Version}
+
+func DefaultContextConfigContents() (contextMap *ContextMap) {
+	ociCliAuth := os.Getenv(OCI_CLI_AUTH_ENV_VAR)
+
+	if ociCliAuth == OCI_CLI_AUTH_INSTANCE_OBO_USER {
```
mrc: nit: could use `switch (ociCliAuth)` (makes it clear it's case analysis, not arbitrary conditional logic)

config/config.go
roroco: So, for the oracle cloudshell and instance principal providers, the default api url (i.e. only if it's not specified in the context) is determined by the provider (it does so by figuring out the region in the environment).
We don't want to write that url onto file, I think, because the fn context file is in the home directory and on cloudshell for e.g., that home directory is shared across regions.

roroco: Added a comment to clarify.

config/config.go
roroco: Added a test for the function

roroco:
Addressed comments.

go.mod
```diff
@@ -1,37 +1,37 @@
 module github.com/fnproject/cli
 
 require (
-	github.com/BurntSushi/toml v0.3.1 // indirect
 	github.com/coreos/go-semver v0.2.0
-	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/fatih/color v0.0.0-20170926111411-5df930a27be2
-	github.com/fnproject/fn_go v0.5.0
+	github.com/fnproject/fn_go v0.5.2-0.20200306235942-b9463d5f4f6b
```
srpurani: Revendor to use latest fn_go with changes to add CS provider.

roroco:
> OracleCSProvider is driven by OCI Auth variable. Would be ideal to separate signing mechanics from CS specific default logic.

I think we can reorganize when there is a case where the token is available outside of a cloudshell environment. But to do so later, we probably need to expand the criteria for when the cloudshell provider is chosen as the default provider i.e. use not just the value of oci_cli_auth but additional environment specific properties.

go.mod
roroco: Done.

srpurani:
> > OracleCSProvider is driven by OCI Auth variable. Would be ideal to separate signing mechanics from CS specific default logic.
> 
> I think we can reorganize when there is a case where the token is available outside of a cloudshell environment. But to do so later, we probably need to expand the criteria for when the cloudshell provider is chosen as the default provider i.e. use not just the value of oci_cli_auth but additional environment specific properties.

Makes sense, we can refactor this when the need arises.

PR #610, opened 2020-04-28
Fix permissions for interpretted function code

riconnon:
Interpretted function code is copied directly in from the filesystem of
the building instance. We don't know what permissions will exist on this
filesystem so we need to ensure after copying it in that the function
user is able to read the code.


mrc:
Do we need to set the umask as well?

riconnon:
> Do we need to set the umask as well?

I don't think so. The umask inside the function isn't the culprit here. Files created inside the function container have correct permissions (hence compiled languages work).
It's the umask of the session in which we created the source files that's the culprit, which we have no control over.

sachin-pikle:
@riconnon is Ruby in the same boat?

riconnon:
Someone pointed out elsewhere that ruby is probably in the same boat.

PR #612, opened 2020-06-24
Enabled Fn CLI to make calls to OCI API. Updated relevant integration tests.

harshpai:
A pull request for an initial code review of the `Adapter` layer code.

This is the initial code review to validate that the code structure and adapter design used are appropriate. This is a Work In Progress and is not ready for merging in master yet. A final code review will be requested once all implementations and tests are completed.


adapter/application.go
```diff
@@ -0,0 +1,35 @@
...
+	// App ID
+	// Read Only: true
+	ID string `json:"id,omitempty"`
+
+	// Name of this app. Must be different than the image name. Can ony contain alphanumeric, -, and _.
+	// Read Only: true
+	Name string `json:"name,omitempty"`
+
+	// A comma separated list of syslog urls to send all function logs to. supports tls, udp or tcp. e.g. tls://logs.papertrailapp.com:1
+	SyslogURL *string `json:"syslog_url,omitempty"`
```
templecloud, line 28: I think this is the right casing. I remember this use to be SyslogUrl? If not done already might want to double check.

adapter/function.go
```diff
@@ -0,0 +1,45 @@
...
+	// Read Only: true
+	// Format: date-time
+	CreatedAt strfmt.DateTime `json:"created_at,omitempty"`
+
+	// Unique identifier
+	// Read Only: true
+	ID string `json:"id,omitempty"`
+
+	// Hot functions idle timeout before container termination. Value in Seconds.
+	IDLETimeout *int32 `json:"idle_timeout,omitempty"`
```
templecloud, line 27: Just out of interest, why is IDLE capitalized?

adapter/interface.go
```diff
@@ -0,0 +1,56 @@
...
+
+type AppClient interface {
+	CreateApp(app *App) (*App, error)
+	GetApp(appName string) (*App, error)
+	UpdateApp(app *App) (*App, error)
+	ListApp(limit int64) ([]*App, error)
+	DeleteApp(appID string) error
+}
+
+type TriggerClient interface {
```
templecloud, line 37: Intentionally left empty as a marker interface?

adapter/oci/app_client.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/spf13/viper"
+	"os"
+)
+
+type AppClient struct {
+	client *functions.FunctionsManagementClient
+}
+
+func (a AppClient) CreateApp(app *adapter.App) (*adapter.App, error) {
+	//TODO: call OCI client
```
templecloud, line 17: Hmmm, this is a bit hard to review without any context. These bits are unimplemented. Is this part of a bigger or multistage strategy for making these changes? If these are really meant to be unimplemented and return null for the short term, then maybe we can some comments explaining the purpose and strategy.

adapter/oss/app_client.go
```diff
@@ -0,0 +1,103 @@
...
+	"os"
+)
+
+type AppClient struct {
+	client *oss.Fn
+}
+
+func (a AppClient) CreateApp(app *adapter.App) (*adapter.App, error) {
+	resp, err := a.client.Apps.CreateApp(&apiapps.CreateAppParams{
+		Context: context.Background(),
```
templecloud, line 19: This looks like a reasonable to create the context.

objects/app/apps.go
```diff
@@ -154,21 +129,21 @@ func CreateApp(a *fnclient.Fn, app *modelsv2.App) (*modelsv2.App, error) {
 		return nil, err
 	}
 
-	fmt.Println("Successfully created app: ", resp.Payload.Name)
-	return resp.Payload, nil
+	fmt.Println("Successfully created app: ", resp.Name)
```
templecloud, line 132: It is probably worth getting the logging story worked out up front. I am new to this CLI code but maybe we could do better? Has this been considered?

adapter/application.go
```diff
@@ -0,0 +1,35 @@
+package adapter
```
templecloud, line 1: Are there new unit tests for this?

commands/deploy.go
```diff
@@ -44,7 +46,8 @@ func DeployCommand() cli.Command {
 }
 
 type deploycmd struct {
-	clientV2 *v2Client.Fn
+	clientV2      *v2Client.Fn
```
ostrain, line 49: Do we still need clientV2? Does apiClientAdapter replace it?

objects/app/apps.go
```diff
@@ -22,11 +23,13 @@ import (
 )
 
 type appsCmd struct {
-	provider provider.Provider
-	client   *fnclient.Fn
+	provider         provider.Provider
+	providerAdapter  adapter.Provider
+	client           *fnclient.Fn
```
ostrain, line 28: Same question here, do we need client or can apiClientAdapter replace it?

adapter/oci/app_client.go
```diff
@@ -0,0 +1,74 @@
...
+
+	if len(resApps) == 0 {
+		fmt.Fprint(os.Stderr, "No apps found\n")
+		return nil, nil
+	}
+
+	return resApps, nil
+}
+
+func convertOCIAppsToAdapterApps(ociApps *[]functions.ApplicationSummary) []*adapter.App {
```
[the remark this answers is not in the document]
harshpai: Yes, that is correct. I'll get to it when I implement GetApp.

adapter/oci/fn_client.go
```diff
@@ -0,0 +1,35 @@
...
+	//TODO: call OCI client
+	return nil, nil
+}
+
+func (f FnClient) DeleteFn(fnID string) error {
+	//TODO: call OCI client
+	return nil
+}
+
+func (f FnClient) ListFn(appID string, limit int64) error {
```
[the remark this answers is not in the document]
harshpai: To be implemented.

adapter/interface.go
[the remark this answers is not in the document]
harshpai: We do know what methods it will have. The same ones used to make manage Triggers in Fn CLI now. This interface will have 2 implementations. An OSS one, that will send calls similar to how we do things now. An OCI one that will be a stub and return an appropriate error response, since we don't support Trigger in OCI Functions.

commands/deploy.go
harshpai: We do need it now since not all references to it have been removed. As the implementation is fleshed out, we will be able to move it from this struct and many others.

objects/app/apps.go
harshpai: Same answer.

adapter/application.go
harshpai: There are none as of now. As the implementation is fleshed out, we will work on the testing strategy that may include unit tests.

objects/app/apps.go
harshpai: This is a great point! However, let's keep in mind that this effort is (narrowly) focussed on enabling the CLI to consume OCI APIs instead of OSS APIs. Keeping in mind time constraints it may be better to track the logging improvements in a separate issue.

adapter/oss/app_client.go
harshpai: This is just a refactor of the code already present. We are not looking to make changes to or introduce calling cancel, configuring timeouts, etc

adapter/oci/app_client.go
harshpai: This is a WIP. The initial round of code review was to get feedback on class design and adapter code structures. As we move to fully implement the "unimplemented bits" things will get in better shape before the final code review.

adapter/function.go
harshpai: // Copied from fn_go/modelsv2/fn.go
// We need to maintain this struct since we have already exposed it to the user

adapter/application.go
harshpai: // Copied from fn_go/modelsv2/app.go
// We need to maintain this struct since we have already exposed it to the user

harshpai:
+ Added ability to FN CLI to send calls to OCI Fn Control Plane APIs. 
- Removed the ability of FN CLI to send calls to OSS Fn Control Plane APIs. 
+ Updated relevant integration tests with a new ability to make calls to OCI API.

adapter/oci/app_client.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	body := functions.CreateApplicationDetails{
+		CompartmentId: &compartmentId,
+		Config:        app.Config,
+		DisplayName:   &app.Name,
+		SubnetIds:     subnetIds,
+	}
+	req := functions.CreateApplicationRequest{CreateApplicationDetails: body,}
+
+	res, err := a.client.CreateApplication(context.Background(), req)
```
harryjsmith, line 40: Would there be a benefit in taking a context as a parameter into these methods, rather than using `context.Background()`? (allow caller to set timeout, cancel etc.).

adapter/oci/app_client.go
```diff
@@ -0,0 +1,208 @@
...
+	} else {
+		return nil, adapter.AppNameNotFoundError{Name: appName}
+	}
+}
+
+func (a AppClient) UpdateApp(app *adapter.App) (*adapter.App, error) {
+	//merge config
+	for k, v := range app.Config {
+		if v == "" {
+			delete(app.Config, k)
```
harryjsmith, line 107: Why do we remove config variables with empty string values here? Is it not valid for a user to set an environment variable with a blank value?

adapter/oci/app_client.go
```diff
@@ -0,0 +1,208 @@
...
+		return nil, err
+	}
+
+	adapterApp := convertOCIAppToAdapterApp(&res.Application)
+	return adapterApp, nil
+}
+
+func (a AppClient) DeleteApp(appID string) error {
+	req := functions.DeleteApplicationRequest{ApplicationId: &appID,}
+	_, err := a.client.DeleteApplication(context.Background(), req)
```
harryjsmith, line 130: Should we be checking responses here (and throughout)?

adapter/oci/app_client.go
```diff
@@ -0,0 +1,208 @@
...
+	for _, ociApp := range *ociApps {
+		app := convertOCIAppSummaryToAdapterApp(&ociApp)
+		resApps = append(resApps, app)
+	}
+
+	return resApps
+}
+
+func convertOCIAppSummaryToAdapterApp(ociApp *functions.ApplicationSummary) *adapter.App {
+	createdAt, _ := strfmt.ParseDateTime(ociApp.TimeCreated.Format(time.RFC3339Nano))
```
harryjsmith, line 175: Should we be checking and handling `err` here?

adapter/oci/fn_client.go
```diff
@@ -0,0 +1,268 @@
...
+
+	digest, err := parseDigestAnnotation(fn)
+	if err != nil {
+		return nil, err
+	}
+
+	//merge config
+	for k, v := range fn.Config {
+		if v == "" {
+			delete(fn.Config, k)
```
harryjsmith, line 107: Same question as for apps - why do we delete keys with blank values?

adapter/oci/provider.go
```diff
@@ -0,0 +1,14 @@
+package oci
+
+import (
+	"github.com/fnproject/cli/adapter"
+	"github.com/oracle/oci-go-sdk/functions"
+)
+
+type Provider struct {
+	FMCClient *functions.FunctionsManagementClient
```
harryjsmith, line 9: Very minor nit: Could this just be `FMClient`? The extra `C` seems redundant

commands/deploy.go
harryjsmith: Yes I suppose we'll want to clean this up before this is merged (to keep things clean).

adapter/application.go
```diff
@@ -0,0 +1,35 @@
...
+
+import "github.com/go-openapi/strfmt"
+
+// Copied from fn_go/modelsv2/app.go
+// We need to maintain this struct since we have already exposed it to the user
+type App struct {
```
harryjsmith, line 7: Thinking out loud here: Is there a reason we want to copy-paste this all, or could we just consume the existing `models` structs from `fn_go`?

oci-test.sh
```diff
@@ -0,0 +1,33 @@
...
+set -ex
+
+export FN_TEST_MODE="OCI"
+
+# FILL IN VALUES BELOW. These values should match with values in the following files:
+# oci-auth/fn/config.yaml
+# oci-auth/fn/contexts/functions-test.yaml
+# oci-auth/oci/config
+# simpleapp/app.json
+export FN_API_URL="https://functions.us-ashburn-1.oraclecloud.com/20181201"
```
harryjsmith, line 11: This domain should be `functions.us-ashburn-1.oci.oraclecloud.com`

test/README.md
```diff
@@ -37,3 +37,34 @@ The test harness runs a specified CLI  binary(either "../fn" or "TEST_CLI_BINARY
 * Don't be spammy : You shouldn't log excessively in tests as this will impact diagnosability when a test fails.  Instead, always log the `CmdResult` you got from the last command that failed - this should include enough diagnostic history to work out what went wrong (including previous commands)
 * Write parallelizable tests: Tests are slow so sequencing them will make the test package slow - the harness includes tools to help make tests isolated (e.g. any app names created with `h.NewFuncName()` will be deleted after a test is done )  - remember to defer `h.Cleanup()` to ensure test state is cleaned up
 * Watch out for the Environment: The CLI package will pass on the surrounding environment to the CLI when its called - (primarily to allow easily overriding local  configuration like FN_API_URL and other env vars)
+
+## Testing against OCI APIs ##
```
harryjsmith, line 41: Could we test this more tightly, with more mocking rather than relying on having to make actual calls?

adapter/oci/app_client.go
ant1eicher: Yup, it is valid to set a config value to empty string. Tested with OCI CLI:

`{
  "data": {
    "compartment-id": "ocid1.compartment.oc1...",
    "config": {
      "foo": ""
    },
    "defined-tags": {},
    "display-name": "faas-demo",
    "freeform-tags": {},
    "id": "ocid1.fnapp.oc1.iad....",
...`

adapter/oci/app_client.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	req := functions.CreateApplicationRequest{CreateApplicationDetails: body,}
+
+	res, err := a.client.CreateApplication(context.Background(), req)
+
+	if err != nil {
+		return nil, err
+	}
+
+	adapterApp := convertOCIAppToAdapterApp(&res.Application)
```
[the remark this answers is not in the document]
harshpai: Based on the Fn examples in the OCI GO SDK, this is the recommended way to do it.
https://github.com/oracle/oci-go-sdk/blob/master/example/example_functions_test.go

adapter/oci/app_client.go
harshpai: We do not provide such facilities (set timeout, cancel etc.) to the caller in our current implementation that consumes Open Source Go CP APIs. Hence, let's keep it the same with the OCI APIs.

adapter/oci/app_client.go
harshpai: Based on the Fn examples in the OCI GO SDK, this is the recommended way to do it.
https://github.com/oracle/oci-go-sdk/blob/master/example/example_functions_test.go#L146

adapter/oci/app_client.go
harshpai: Nice catch! Fixed it. Thanks.

adapter/oci/provider.go
harshpai: Fixed.

adapter/application.go
harshpai: We would like to be able to eventually retire the `fn_go` library dependency form this Fn CLI. This change allows us to take a step in that direction.

oci-test.sh
harshpai: Will ask the user to fill this in.

adapter/interface.go
```diff
@@ -0,0 +1,69 @@
...
+// NameNotFoundError error for function not found when looked up by name
+type FunctionNameNotFoundError struct {
+	Name string
+}
+
+func (n FunctionNameNotFoundError) Error() string {
+	return fmt.Sprintf("function %s not found", n.Name)
+}
+
+// NameNotFoundError error for trigger not found when looked up by name
```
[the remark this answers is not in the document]
harshpai: Fixed

test/README.md
harshpai: The current integration tests provide end to end tests against real production OCI API endpoints. This should provide more than sufficient test coverage for this change.

harshpai:
Work In Progress on merge config.

test/README.md
harryjsmith: Yeah, my only concern here is that we can't automate that through CI easily

harryjsmith:
Superseded by #616

PR #613, opened 2020-08-20
Add Python 3.8.5 and make Python images smaller

mrc:
## Python 3.8.5 is a supported language, and the default for "python"

It's the default for "python", because it leads to smaller images.

## Python images are now smaller

This patch changes the file mode of the `/python` tree as part of the
build image, which avoids duplicating the pip installed libraries in
the final layer. This makes docker push and pull much faster.

"Hello world" with Python 3.8.5:

| requirements.txt | image size now | image size before | saving |
| ---------------- | -------------- | ----------------- | ------ |
| fdk              |          136MB |             161MB |    15% |
| fdk & oci        |          202MB |             292MB |    30% |

## Minor fixups

* Make Python boilerplate meet PEP 8 standard
* Remove unused func "dockerBuildError" (dead code)

### DIFF
--- a/langs/base.go
+++ b/langs/base.go
@@ -2,7 +2,6 @@ package langs
 
 import (
 	"errors"
-	"fmt"
 	"os"
 )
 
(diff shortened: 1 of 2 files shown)


reclaro:
Your table shows a great improvement, have you tried to verify that with this version of the code, the creation of a python function works as expected? If so I think we are good to go.

reclaro (approved):
LGTM

PR #614, opened 2020-10-02
Oracle CS contexts: retrieve domain from IMDS

harryjsmith:
This will need revendoring with https://github.com/fnproject/fn_go/pull/40 once merged before this PR is merged

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -55,26 +55,6 @@ func (*ViperConfigSource) IsSet(key string) bool {
 	return viper.IsSet(key)
 }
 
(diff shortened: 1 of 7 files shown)


config/config.go
```diff
@@ -55,26 +55,6 @@ func (*ViperConfigSource) IsSet(key string) bool {
 	return viper.IsSet(key)
 }
 
-func DefaultOracleCSRegionContextContents() (contextMap *ContextMap, err error) {
```
harryjsmith, line 58: This is duplicate code that is unused as far as I can tell

roroco (approved):

PR #616, opened 2020-12-08
Revendor fn_go: switch Oracle providers to OCI API

harryjsmith:
See https://github.com/fnproject/fn_go/pull/42

Oracle providers will now use OCI API under-the-hood.

Note: OCI_GO_SDK_DEBUG=v must be used instead of DEBUG=1 to dump request/responses when an Oracle provider is in use.


ant1eicher (approved):
LGTM.

PR #624, opened 2021-04-13
Migrate to maven central

roneetshaw:
CLI changes for reading the Latest version from Maven central API and uses Bintray API as a fallback


bgputtan (approved):

langs/java.go
```diff
@@ -174,8 +177,15 @@ func mavenOpts() string {
 /*    TODO temporarily generate maven project boilerplate from hardcoded values.
 Will eventually move to using a maven archetype.
 */
-func pomFileContent(APIversion, javaVersion string) string {
-	return fmt.Sprintf(pomFile, APIversion, javaVersion, javaVersion)
+func pomFileContent(APIversion, javaVersion, groupId, pomType string) string {
+	if groupId == "io.fnproject.com" {
```
hhexo, line 181: Could this be `if groupId == "io.fnproject.com" || pomType == "maven"`?
The `Sprintf` seems identical in both cases.

langs/kotlin.go
```diff
@@ -163,8 +166,15 @@ func kotlinMavenOpts() string {
 
 /*    TODO temporarily generate maven project boilerplate from hardcoded values.
       Will eventually move to using a maven archetype.*/
-func kotlinPomFileContent(APIversion string) string {
-	return fmt.Sprintf(kotlinPomFile, APIversion)
+func kotlinPomFileContent(APIversion, groupId, pomType string) string {
+	if groupId == "io.fnproject.com" {
```
hhexo, line 170: Similarly I think this case can be merged with the pomType == maven case.

langs/java.go
```diff
@@ -184,20 +194,71 @@ func (h *JavaLangHelper) getFDKAPIVersion() (string, error) {
...
+			h.pomType = "bintray"
+			return version, e1
+		}
+	}
+
+	//In all other case return error as latest FDK version is not identified
+	return "", err
+}
+
+func (h *JavaLangHelper) getURLResponse(url string) (*bytes.Buffer, error) {
```
hhexo, line 259: As the problem with verifying certs is only there for bintray, can we make sure that we don't use insecureSkipVerify with Maven Central? We definitely want to make sure packages downloaded from there are genuine. We could pass a flag to this function and only set it to true for the bintray path.

langs/kotlin.go
roneetshaw: Done

langs/java.go
roneetshaw: Done

langs/java.go
roneetshaw: Done

hhexo (approved):

langs/java.go
```diff
@@ -184,50 +190,121 @@ func (h *JavaLangHelper) getFDKAPIVersion() (string, error) {
...
+	buf, err = h.getURLResponse(comURL, false)
+	if err == nil {
+		version, e1 := h.parseMavenResponse(*buf)
+		if e1 == nil {
+			h.pomType = "maven"
+			return version, e1
+		}
+	}
+
+	//Second time search for com.fnproject.fn from Bintray to get the latest version, if both com.fnproject.fn and io.fnproject.fn fails
```
hhexo, line 227: The comment still mentions io.fnproject here

langs/kotlin.go
```diff
@@ -173,50 +179,120 @@ func (lh *KotlinLangHelper) getFDKAPIVersion() (string, error) {
...
+	buf, err = lh.getURLResponse(comURL, false)
+	if err == nil {
+		version, e1 := lh.parseMavenResponse(*buf)
+		if e1 == nil {
+			lh.pomType = "maven"
+			return version, e1
+		}
+	}
+
+	//Second time search for com.fnproject.fn from Bintray to get the latest version, if both com.fnproject.fn and io.fnproject.fn fails
```
hhexo, line 216: Similarly this comment still references io.fnproject

hhexo (approved):
Approved conditional on fixing a couple of comments.

hhexo (approved):

langs/java.go
```diff
@@ -184,50 +190,121 @@ func (h *JavaLangHelper) getFDKAPIVersion() (string, error) {
...
+		version, e1 := h.parseBintrayResponse(*buf)
+		if e1 == nil {
+			h.pomType = "bintray"
+			return version, e1
+		}
+	}
+
+	//In all other case return error as latest FDK version is not identified
+	return "", err
+}
```
metamemelord, line 239: Consider accepting variadic params of structs containing type & url and loop over it? It will prevent code duplication.

langs/java.go
```diff
@@ -184,50 +190,121 @@ func (h *JavaLangHelper) getFDKAPIVersion() (string, error) {
...
-	resp, err := client.Get(versionURL)
+	resp, err := client.Get(url)
 	if err != nil || resp.StatusCode != 200 {
-		return "", fetchError
+		return nil, fmt.Errorf("Failed to fetch response from URL %s Error: %v Status: %d", url, err, resp.StatusCode)
 	}
-
-	buf := bytes.Buffer{}
+	buf := &bytes.Buffer{}
 	_, err = buf.ReadFrom(resp.Body)
```
metamemelord, line 259: Close the response body after consuming.

langs/java.go
```diff
@@ -184,50 +190,121 @@ func (h *JavaLangHelper) getFDKAPIVersion() (string, error) {
...
-
-	resp, err := client.Get(versionURL)
+	resp, err := client.Get(url)
 	if err != nil || resp.StatusCode != 200 {
-		return "", fetchError
+		return nil, fmt.Errorf("Failed to fetch response from URL %s Error: %v Status: %d", url, err, resp.StatusCode)
 	}
-
-	buf := bytes.Buffer{}
+	buf := &bytes.Buffer{}
```
metamemelord, line 258: Consider converting this into a byte array instead of a buffer? We have a `io.Reader` (resp.Body) here and we are converting it into bytes.Buffer and then to Bytes in child functions.

langs/kotlin.go
```diff
@@ -173,50 +179,120 @@ func (lh *KotlinLangHelper) getFDKAPIVersion() (string, error) {
...
+		version, e1 := lh.parseBintrayResponse(*buf)
+		if e1 == nil {
+			lh.pomType = "bintray"
+			return version, e1
+		}
+	}
+
+	//In all other case return error as latest FDK version is not identified
+	return "", err
+}
```
metamemelord, line 228: I see a lot of couple of common methods in both helpers. They are in same package, perhaps we should make them unexported functions in the package and place in Java helper file.

langs/java.go
roneetshaw: Done

langs/java.go
roneetshaw: DOne

langs/java.go
metamemelord: Since we are never likely to have dynamic number of package repos, this change is good.

langs/kotlin.go
roneetshaw: DOne

metamemelord (approved):
LGTM.

PR #627, opened 2021-05-31
Runtime updates with version stamping

ssunny3141:


common/common.go
```diff
@@ -123,6 +132,97 @@ func BuildFuncV20180708(verbose bool, fpath string, funcfile *FuncFileV20180708,
...
 }
 
+func imageStampFuncFile(fpath string, funcfile *FuncFile) (*FuncFile, error) {
+
+	dir := filepath.Dir(fpath)
+	dockerfile := filepath.Join(dir, "Dockerfile")
+
+	// detect if build and run image both are absent and runtime is not docker then update them
+	fmt.Println(funcfile.BuildImage)
+	fmt.Println(funcfile.RunImage)
```
roneetshaw, line 142: Are these print statements required?

ssunny3141: Will remove them

langs/python.go
```diff
@@ -32,7 +34,14 @@ func (h *PythonLangHelper) GenerateBoilerplate(path string) error {
 		return err
 	}
 	depFile := "requirements.txt"
-	if err := ioutil.WriteFile(depFile, []byte(reqsPythonSrcBoilerplate), os.FileMode(0644)); err != nil {
+	fdkVersionInfo, err := h.GetLatestFDKVersion()
+	if err != nil {
+		fmt.Println("Unable to get latest FDK version, using default")
```
sudhaku, line 39: Where are we setting fdkVersionInfo to default if err was not nil ?

common/common.go
```diff
@@ -123,6 +132,95 @@ func BuildFuncV20180708(verbose bool, fpath string, funcfile *FuncFileV20180708,
 	return funcfile, nil
 }
 
+func imageStampFuncFile(fpath string, funcfile *FuncFile) (*FuncFile, error) {
+
+	dir := filepath.Dir(fpath)
+	dockerfile := filepath.Join(dir, "Dockerfile")
+
+	// detect if build and run image both are absent and runtime is not docker then update them
+	if !Exists(dockerfile) && funcfile.Runtime != FuncfileDockerRuntime && funcfile.BuildImage == "" && funcfile.RunImage == "" {
```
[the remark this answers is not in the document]
ssunny3141: This is to ensure that for runtime 'docker' we don't stamp versions.
fn init --runtime docker 
This check is there at other places too wherever dev/runtime image were getting picked.

langs/python.go
ssunny3141: @metamemelord

common/common.go
```diff
@@ -123,6 +132,95 @@ func BuildFuncV20180708(verbose bool, fpath string, funcfile *FuncFileV20180708,
...
+	if !Exists(dockerfile) && funcfile.Runtime != FuncfileDockerRuntime && funcfile.Build_image == "" && funcfile.Run_image == "" {
+
+		helper := langs.GetLangHelper(funcfile.Runtime)
+
+		// get the lang without any version
+		langRuntime := helper.Runtime()
+
+		// if fallback is not supported then continue with same lang helper
+		if langs.IsFallbackSupported(langRuntime) {
+			helper = langs.GetFallbackLangHelper(langRuntime)
```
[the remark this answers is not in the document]
ssunny3141: Latest runtime if initialized using newer version (this version) of cli will always have build and runtime image flag set, so check on outer condition will fail and won't fallback.
This check is only func.yaml file where build and runtime images are not not.

common/common.go
```diff
@@ -123,6 +132,95 @@ func BuildFuncV20180708(verbose bool, fpath string, funcfile *FuncFileV20180708,
...
+			return funcfile, err
+		}
+		funcfile.BuildImage = bi
+
+
+		// In case of multistage build there will be a runtime image
+		if helper.IsMultiStage() {
+			ri, err := helper.RunFromImage()
+			if err != nil {
+				return funcfile, nil
```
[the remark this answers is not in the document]
ssunny3141: Good catch! will change.

langs/python.go
metamemelord: We go via the current flow: assign `fdk` in requirements.txt and not add version. On the other hand, if we do get the latest version, we add `fdk>=<version>` in requirements.txt.

metamemelord: Failure to fetch latest FDK version can mean one of two things:

1. Pypi is down, in this case user function will not deploy - this is sort unrecoverable at the moment.
2. Pypi changes response json. In this case, we have 2 choices: prevent user to deploy function altogether until we fix the parsing logic OR leave the fdk tag empty in requirements.txt and let pip pick latest automatically. Current flow takes the latter path.

metamemelord (changes requested):
Please squash the commits. LGTM otherwise.

roneetshaw:
Changes look good.

roneetshaw (approved):

common/funcfile.go
```diff
@@ -234,6 +234,8 @@ func encodeFuncfileJSON(path string, ff *FuncFile) error {
 
 // EncodeFuncfileYAML encodes function file.
 func EncodeFuncfileYAML(path string, ff *FuncFile) error {
+	fmt.Println(ff.BuildImage)
+	fmt.Println(ff.RunImage)
```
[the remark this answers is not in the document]
ssunny3141: Removed.

Issue #629, opened 2021-08-24
Support webpack dist


Joelith:
It would be great if the fn cli could support webpack compiled dist files instead of having to send the entire code base and node_modules in with the deployment. Not only would this reduce file size, it would also allow us to add test code that we don't want to submit to the server. For a project I'm working on I have a lot of test scripts and dummy data, many of which are Mbs in size. If I didn't have to have those included it would be great. 

I've worked on a patch to fix this and will create a PR shortly.

PR #630, opened 2021-08-24
Added webpack support for Node

Joelith:
Added the ability to use a compiled dist folder script rather than the full code. Reduces overall file size and allows you to have test code that isn't compiled and sent to the server. 

Resolves Issue #629

### DIFF
--- a/config/version.go
+++ b/config/version.go
@@ -9,7 +9,7 @@ import (
 )
 
 // Version of Fn CLI
-var Version = "0.6.8"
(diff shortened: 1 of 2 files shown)


Joelith:
I should add this code assumes you are compiling to dist/main.js (which is the default). You will need to install webpack (`npm install webpack webpack-cli --dev`) and then add a webpack.config.js like so:

```
module.exports = {
  entry: './func.js',
  target: 'node'
};
```

PR #637, opened 2021-12-06
Added support for dotnet

metamemelord:
Added support for Dotnet

### DIFF
--- a/langs/base.go
+++ b/langs/base.go
@@ -17,8 +17,13 @@
 package langs
 
 import (
+	"encoding/json"
 	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
 	"os"
+	"path/filepath"
 )
 
 // not a map because some helpers can handle multiple keys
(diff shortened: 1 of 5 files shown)


langs/csharp.go
```diff
@@ -0,0 +1,170 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"path/filepath"
+)
+
+type CSharpHelper struct {
```
Preeti-Misra, line 29: Should the CSharpHelper struct be renamed to CSharpLangHelper just to be consistent with the other language helpers structures like go, java etc ?

Preeti-Misra (commented):
Tested the cli changes for csharp and they look good except one minor comment regarding the name of the language helper struct for CSharp.

langs/csharp.go
metamemelord: Yep, done.

Preeti-Misra (approved):

PR #640, opened 2022-01-03
Podman has been integrated to be used as an optional replacement for docker

rishabhsri18:
Changes which will pick the container engine name from the environment and replaced all the docker argument with the variable of containerEngineType

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -114,7 +114,7 @@ func BuildFunc(verbose bool, fpath string, funcfile *FuncFile, buildArg []string
 		return nil, err
 	}
 
-	if err := dockerBuild(verbose, fpath, funcfile, buildArg, noCache); err != nil {
+	if err := containerEngineBuild(verbose, fpath, funcfile, buildArg, noCache); err != nil {
 		return nil, err
 	}
 
@@ -141,7 +141,7 @@ func BuildFuncV20180708(verbose bool, fpath string, funcfile *FuncFileV20180708,
 		return nil, err
 	}
 
-	if err := dockerBuildV20180708(verbose, fpath, funcfile, buildArg, noCache); err != nil {
+	if err := containerEngineBuildV20180708(verbose, fpath, funcfile, buildArg, noCache); err != nil {
 		return nil, err
 	}
 
@@ -177,7 +177,6 @@ func imageStampFuncFile(fpath string, funcfile *FuncFile) (*FuncFile, error) {
 			return funcfile, err
 		}
 		funcfile.BuildImage = bi
-
 
 		// In case of multistage build there will be a runtime image
 		if helper.IsMultiStage() {
@@ -274,8 +273,26 @@ func PrintContextualInfo() {
 	fmt.Println("Current Context: ", currentContext)
 }
 
-func dockerBuild(verbose bool, fpath string, ff *FuncFile, buildArgs []string, noCache bool) error {
-	err := dockerVersionCheck()
+func GetContainerEngineType() (string, error) {
+	containerEngineType := viper.GetString(config.ContainerEngineType)
+	err := config.ValidateContainerEngineType(containerEngineType)
+	if err != nil {
+		return "", err
+	}
+	if containerEngineType == "" {
+		return "docker", nil
+	}
+	return containerEngineType, nil
+}
+
+func containerEngineBuild(verbose bool, fpath string, ff *FuncFile, buildArgs []string, noCache bool) error {
+	containerEngineType, err := GetContainerEngineType()
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("Using Container engine", containerEngineType)
+	err = containerEngineVersionCheck(containerEngineType)
 	if err != nil {
 		return err
 	}
@@ -305,7 +322,7 @@ func dockerBuild(verbose bool, fpath string, ff *FuncFile, buildArgs []string, n
 			}
 		}
 	}
-	err = RunBuild(verbose, dir, ff.ImageName(), dockerfile, buildArgs, noCache)
+	err = RunBuild(verbose, dir, ff.ImageName(), dockerfile, buildArgs, noCache, containerEngineType)
 	if err != nil {
 		return err
 	}
@@ -319,8 +336,14 @@ func dockerBuild(verbose bool, fpath string, ff *FuncFile, buildArgs []string, n
 	return nil
 }
 
-func dockerBuildV20180708(verbose bool, fpath string, ff *FuncFileV20180708, buildArgs []string, noCache bool) error {
-	err := dockerVersionCheck()
+func containerEngineBuildV20180708(verbose bool, fpath string, ff *FuncFileV20180708, buildArgs []string, noCache bool) error {
+	containerEngineType, err := GetContainerEngineType()
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("Using Container engine", containerEngineType)
+	err = containerEngineVersionCheck(containerEngineType)
 	if err != nil {
 		return err
 	}
@@ -350,7 +373,7 @@ func dockerBuildV20180708(verbose bool, fpath string, ff *FuncFileV20180708, bui
 			}
 		}
 	}
-	err = RunBuild(verbose, dir, ff.ImageNameV20180708(), dockerfile, buildArgs, noCache)
+	err = RunBuild(verbose, dir, ff.ImageNameV20180708(), dockerfile, buildArgs, noCache, containerEngineType)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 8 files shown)


common/common.go
```diff
@@ -634,12 +665,16 @@ func ExtractConfig(configs []string) map[string]string {
 
 // DockerPush pushes to docker registry.
 func DockerPush(ff *FuncFile) error {
```
theatul, line 667: we need to modify these function names.

common/common.go
```diff
@@ -274,8 +273,24 @@ func PrintContextualInfo() {
 	fmt.Println("Current Context: ", currentContext)
 }
 
+func GetContainerEngineType() (string, error) {
```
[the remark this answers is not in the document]
rishabhsri18: We set docker as default if no info is present it's just the reference where we get the config details from viper. I will return docker here as well if we can't find any value

common/common.go
```diff
@@ -442,17 +470,20 @@ func RunBuild(verbose bool, dir, imageName, dockerfile string, buildArgs []strin
 	return nil
 }
 
-func dockerVersionCheck() error {
-	out, err := exec.Command("docker", "version", "--format", "{{.Server.Version}}").Output()
+func dockerVersionCheck(containerEngineType string) error {
```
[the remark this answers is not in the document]
rishabhsri18: It check the container engine w.r.t the version call, we have a minimum version requirement for docker and not for podman. So just checked the podman version and docker condition remains the same

config/config.go
```diff
@@ -247,6 +259,14 @@ func LoadConfiguration(c *cli.Context) error {
 	return nil
 }
 
+func ValidateContainerEngineType(containerEngineType string) error {
+	if strings.Compare(containerEngineType, "docker") != 0 && strings.Compare(containerEngineType, "podman") != 0 &&
+		strings.Compare(containerEngineType, "") != 0 {
+		return errors.New("Invalid Container Engine")
+	}
+	return nil
```
metamemelord, line 267: I think a switch with fallthrough will suffice.

common/common.go
```diff
@@ -632,34 +664,43 @@ func ExtractConfig(configs []string) map[string]string {
...
+		return fmt.Errorf("error running docker/podman push, are you logged into docker?: %v", err)
 	}
 	return nil
 }
 
-// DockerPush pushes to docker registry.
-func DockerPushV20180708(ff *FuncFileV20180708) error {
-	err := ValidateFullImageName(ff.ImageNameV20180708())
+// ContainerEnginePush pushes to docker registry.
+func ContainerEnginePushV20180708(ff *FuncFileV20180708) error {
```
metamemelord, line 688: "image registry".

Can be renamed `PushImageV20180708` to be explicit.

common/common.go
```diff
@@ -632,34 +664,43 @@ func ExtractConfig(configs []string) map[string]string {
 	return c
 }
 
-// DockerPush pushes to docker registry.
-func DockerPush(ff *FuncFile) error {
-	err := ValidateFullImageName(ff.ImageName())
+// ContainerEnginePush pushes to docker registry.
+func ContainerEnginePush(ff *FuncFile) error {
```
metamemelord, line 668: Same as last.

common/common.go
```diff
@@ -319,8 +336,14 @@ func dockerBuild(verbose bool, fpath string, ff *FuncFile, buildArgs []string, n
...
 
-func dockerBuildV20180708(verbose bool, fpath string, ff *FuncFileV20180708, buildArgs []string, noCache bool) error {
-	err := dockerVersionCheck()
+func containerEngineBuildV20180708(verbose bool, fpath string, ff *FuncFileV20180708, buildArgs []string, noCache bool) error {
+	containerEngineType, err := GetContainerEngineType()
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("Using Container engine ", containerEngineType)
```
metamemelord, line 345: More like an OCD 😄 commas between `Println` add spaced automatically. I hope all these will be cleaned up before merging.

common/common.go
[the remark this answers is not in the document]
metamemelord: We can use `viper.SetDefault` to ensure `""` is never returned.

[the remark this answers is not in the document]
rishabhsri18: So the viper.SetDefault is there which is setting up the default value for the containerEngine value to docker as default

common/common.go
```diff
@@ -632,34 +664,43 @@ func ExtractConfig(configs []string) map[string]string {
...
 		return err
 	}
 	fmt.Printf("Pushing %v to docker registry...", ff.ImageName())
-	cmd := exec.Command("docker", "push", ff.ImageName())
+	cmd := exec.Command(containerEngineType, "push", ff.ImageName())
 	cmd.Stderr = os.Stderr
 	cmd.Stdout = os.Stdout
 	if err := cmd.Run(); err != nil {
-		return fmt.Errorf("error running docker push, are you logged into docker?: %v", err)
+		return fmt.Errorf("error running docker/podman push, are you logged into docker?: %v", err)
```
roneetshaw, line 682: Can we pass the containerEngineType here?
Remove docker from the statement

common/common.go
```diff
@@ -632,34 +664,43 @@ func ExtractConfig(configs []string) map[string]string {
...
 	}
+	fmt.Println("Using Container engine ", containerEngineType, "to push")
 	fmt.Printf("Pushing %v to docker registry...", ff.ImageNameV20180708())
-	cmd := exec.Command("docker", "push", ff.ImageNameV20180708())
+	cmd := exec.Command(containerEngineType, "push", ff.ImageNameV20180708())
 	cmd.Stderr = os.Stderr
 	cmd.Stdout = os.Stdout
 	if err := cmd.Run(); err != nil {
-		return fmt.Errorf("error running docker push, are you logged into docker?: %v", err)
+		return fmt.Errorf("error running docker/podman push, are you logged into docker?: %v", err)
```
roneetshaw, line 703: Can we pass the containerEngineType here?
Remove docker from the statement

common/common.go
```diff
@@ -886,15 +927,19 @@ func ListTriggersInFunc(c *cli.Context, client *fnclient.Fn, fn *modelsv2.Fn) ([
...
+	}
 	fmt.Println("Running init-image: " + initImage)
 
 	args := []string{"run", "--rm", "-e", "FN_FUNCTION_NAME=" + fName}
 	args = append(args, proxyArgs()...)
 	args = append(args, initImage)
 
-	fmt.Printf("Executing docker command: %s\n", strings.Join(args, " "))
-	cmd := exec.Command("docker", args...)
+	fmt.Printf("Executing docker/podman command: %s\n", strings.Join(args, " "))
```
roneetshaw, line 941: Pass the ```containerEngineType``` in the info message

common/common.go
```diff
@@ -886,15 +927,19 @@ func ListTriggersInFunc(c *cli.Context, client *fnclient.Fn, fn *modelsv2.Fn) ([
 	return resTriggers, nil
 }
 
-func DockerRunInitImage(initImage string, fName string) error {
+func RunInitImage(initImage string, fName string) error {
+	containerEngineType, err := GetContainerEngineType()
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
rishabhsri18: So this is basically required as we don't have any repetitive code it is different for every commands and i think for getting the data in every command we need to call this function

common/common.go
```diff
@@ -442,17 +471,20 @@ func RunBuild(verbose bool, dir, imageName, dockerfile string, buildArgs []strin
...
 	}
 	// dev / test builds append '-ce', trim this
 	trimmed := strings.TrimRightFunc(string(out), func(r rune) bool { return r != '.' && !unicode.IsDigit(r) })
 
 	v, err := semver.NewVersion(trimmed)
 	if err != nil {
-		return fmt.Errorf("could not check Docker version: %v", err)
+		return fmt.Errorf("could not check Docker/Podman version: %v", err)
+	}
+	if containerEngineType == "podman" {
```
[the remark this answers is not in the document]
rishabhsri18: We can definitely use some factory pattern design to get rid of multiple checking but we need to refactor a lot of code for this

common/common.go
```diff
@@ -141,7 +141,7 @@ func BuildFuncV20180708(verbose bool, fpath string, funcfile *FuncFileV20180708,
 		return nil, err
 	}
 
-	if err := dockerBuildV20180708(verbose, fpath, funcfile, buildArg, noCache); err != nil {
+	if err := containerEngineBuildV20180708(verbose, fpath, funcfile, buildArg, noCache); err != nil {
```
theatul, line 144: We can probably rename this ti just Build<version> or a more appropriate name in a future change.

metamemelord (approved):
LGTM.

roneetshaw (approved):

PR #642, opened 2022-01-17
FAAS-7483: Support creating OCIR repos in the non-root compartment.

parthasarathy-menon:
Fn context, oracle.image-compartment-id is used to specify the
compartment id to which the function image is to be deployed to.
If the the repository already exists and oracle.image-compartment-id is
specified, the deployment will fail is the repository is not present
same compartment id as the image-compartment-id.


commands/deploy.go
```diff
@@ -343,9 +343,48 @@ func (p *deploycmd) deployFuncV20180708(c *cli.Context, app *models.App, funcfil
...
+
+		// Create artifactsClient and set region.
+		artifactsClient, err := artifacts.NewArtifactsClientWithConfigurationProvider(oracleProvider.ConfigurationProvider)
+		if err != nil {
+			return err
+		}
+		artifactsClient.SetRegion(getRegion(oracleProvider))
+
+		// Create the repository in ImageCompartmentID
+		repositoryExists, err := createContainerRepositoryInCompartment(repositoryName, oracleProvider.ImageCompartmentID, artifactsClient)
```
ahjones, line 369: I wonder if the logic in this area might be more clear if it was switched around slightly?

What do you think about checking for existence of the container first, with `doesRepositoryExistInComparment`. If that's true, then you're done. If false, then proceed to create the repository, and if that fails, you know it's a failure that you can't continue with.

That way the response from `createContainerRepositoryInCompartment` doesn't have to distinguish between 409s and other responses.

commands/deploy.go
```diff
@@ -697,3 +736,44 @@ func findMissingValues(signingDetails common.SigningDetails) string {
...
+	return repositoryNamePrefix[:len(repositoryNamePrefix)-1], nil
+}
+
+// Checks if the repostitory exists in the compartment
+func doesRepositoryExistInCompartment(repositoryName string, compartmentID string, artifactsClient artifacts.ArtifactsClient) (bool, error) {
+	response, err := artifactsClient.ListContainerRepositories(context.Background(), artifacts.ListContainerRepositoriesRequest{
+		CompartmentId: &compartmentID,
+		DisplayName:   &repositoryName})
+	if err != nil {
+		return false, fmt.Errorf("failed to lookup Container Repository due to %s", err)
```
ahjones, line 758:
```suggestion
		return false, fmt.Errorf("failed to lookup Container Repository due to %w", err)
```

commands/deploy.go
parthasarathy-menon: Nice catch. Updating it that way

commands/deploy.go
```diff
@@ -343,9 +343,42 @@ func (p *deploycmd) deployFuncV20180708(c *cli.Context, app *models.App, funcfil
...
+		artifactsClient.SetRegion(getRegion(oracleProvider))
+
+		// Check if the repository exists in ImageCompartmentID
+		repositoryExists, err := doesRepositoryExistInCompartment(repositoryName, oracleProvider.ImageCompartmentID, artifactsClient)
+		if err != nil {
+			return err
+		}
+		if !repositoryExists {
+			createContainerRepositoryInCompartment(repositoryName, oracleProvider.ImageCompartmentID, artifactsClient)
+			if err != nil {
```
ahjones, line 375: This looks wrong. `err` hasn't been assigned in the inner block.

commands/deploy.go
```diff
@@ -343,9 +343,42 @@ func (p *deploycmd) deployFuncV20180708(c *cli.Context, app *models.App, funcfil
...
 		funcfile.Name = filepath.Base(filepath.Dir(funcfilePath)) // todo: should probably make a copy of ff before changing it
 	}
-	fmt.Printf("Deploying %s to app: %s\n", funcfile.Name, app.Name)
 
-	var err error
+	oracleProvider, _ := getOracleProvider()
+	if oracleProvider != nil && oracleProvider.ImageCompartmentID != "" {
+		// If the provider is Oracle and ImageCompartmentID is present, we need to deploy image to the ImageCompartmentID.
+
+		// Check Registry and build the repository name.
```
ahjones, line 351: nit: most of these small comments aren't doing a lot for me. e.g., here "check registry and build the repository name" just says exactly what the code is doing. Personally I think it would be better (but not required) to write a small comment at the top describing roughly /why/ things are happening the way they are. e.g., "A repository must have a unique name within the tenancy. If one doesn't exist create it in the compartment; if one exists already check to see if it is in the right compartment. If it's not in the right compartment we can't continue".

commands/deploy.go
parthasarathy-menon: My bad. This code has moved around so much that I am missing things.

commands/deploy.go
parthasarathy-menon: Here the comments I am leaving are like steps and the code is what the step does. Its much easier to understand what is happening later on when reading than to see a chunk of code and figure out what it does, especially for someone new. I'll add why we do it at the top and space it so it helps in a similar way.

client/api.go
```diff
@@ -26,3 +26,7 @@ import (
 func CurrentProvider() (provider.Provider, error) {
 	return fn_go.DefaultProviders.ProviderFromConfig(viper.GetString(config.ContextProvider), &config.ViperConfigSource{}, &provider.TerminalPassPhraseSource{})
 }
+
+func Registry() string {
```
ahjones, line 30: I don't think that `Registry` should be added to `api.go`.

commands/deploy.go
```diff
@@ -343,9 +343,41 @@ func (p *deploycmd) deployFuncV20180708(c *cli.Context, app *models.App, funcfil
...
+	oracleProvider, _ := getOracleProvider()
+	if oracleProvider != nil && oracleProvider.ImageCompartmentID != "" {
+		// If the provider is Oracle and ImageCompartmentID is present, we need to deploy image to the ImageCompartmentID.
+		// The repository name should be unique throughout a tenancy. We check if a repository exists in the compartment and create it if it doesn't already exist.
+		// If the creation fails, it could be because the repository name aready exists in a different compartment.
+
+		if client.Registry() == "" {
+			return errors.New("registry name must be set, be sure to configure your context file")
+		}
+		res, err := getRepositoryNamePrefix(client.Registry())
```
ahjones, line 356: I think that the existing function `getRepositoryName` will do what you want here. Additionally it'll support `name` containing a different registry
```suggestion
		res, err := getRepositoryName(funcfile)
```

commands/deploy.go
```diff
@@ -697,3 +729,36 @@ func findMissingValues(signingDetails common.SigningDetails) string {
 	}
 	return strings.Join(missingValues, ",")
 }
+
+func getRepositoryNamePrefix(registry string) (string, error) {
```
ahjones, line 733: I'm not sure that this fn is needed, if you use `getRepositoryName` as suggested elsewhere.

commands/deploy.go
```diff
@@ -697,3 +729,36 @@ func findMissingValues(signingDetails common.SigningDetails) string {
...
+	return strings.Join(parts[2:], "/"), nil
+}
+
+// Checks if the repostitory exists in the compartment
+func doesRepositoryExistInCompartment(repositoryName string, compartmentID string, artifactsClient artifacts.ArtifactsClient) (bool, error) {
+	response, err := artifactsClient.ListContainerRepositories(context.Background(), artifacts.ListContainerRepositoriesRequest{
+		CompartmentId: &compartmentID,
+		DisplayName:   &repositoryName})
+	if err != nil {
+		return false, fmt.Errorf("failed to lookup Container Repository due to %w", err)
```
ahjones, line 747:
```suggestion
		return false, fmt.Errorf("failed to lookup container repository due to %w", err)
```

commands/deploy.go
```diff
@@ -697,3 +729,36 @@ func findMissingValues(signingDetails common.SigningDetails) string {
...
+	}
+	if *response.RepositoryCount == 1 {
```
ahjones, line 749: I'm not sure about this suggestion, but it seems that if the count is bigger than one it exists.
```suggestion
	if *response.RepositoryCount >= 1 {
```

ahjones (approved):

riconnon (approved):

PR #645, opened 2022-02-23
Added support for java version 17 in cli

rishabhsri18:


langs/java.go
```diff
@@ -81,6 +83,8 @@ func (h *JavaLangHelper) RunFromImage() (string, error) {
 		return fmt.Sprintf("fnproject/fn-java-fdk:%s", fdkVersion), nil
 	} else if h.version == "11" {
 		return fmt.Sprintf("fnproject/fn-java-fdk:jre11-%s", fdkVersion), nil
+	} else if h.version == "17" {
+		return fmt.Sprintf("fnproject/fn-java-fdk:jre17-%s", fdkVersion), nil
```
denismakogon, line 87: that's a good time to stop naming something as JRE, this concept is no longer valid for Java since JDK 11.

rishabhsri18: What can it named as we are having a concept of creating two images. One for the build image and other for the run image so it is not exactly the jre but the run image for the fdk

rishabhsri18: Thanks for your suggestion. We are going to pick this up in future https://jira.oci.oraclecorp.com/browse/FAAS-7797

theatul (approved):

PR #652, opened 2022-05-31
Added support for net6.0

metamemelord:
Added support for net6.0

### DIFF
--- a/test/cli_lang_boilerplate_test.go
+++ b/test/cli_lang_boilerplate_test.go
@@ -29,6 +29,8 @@ var Runtimes = []struct {
 }{
 	{"go", ""},
 	{"dotnet", ""},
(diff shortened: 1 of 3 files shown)


test/cli_lang_boilerplate_test.go
```diff
@@ -29,6 +29,8 @@ var Runtimes = []struct {
 }{
 	{"go", ""},
 	{"dotnet", ""},
+	{"dotnet6.0", ""},
```
VikashLNU, line 32: can you reverse this order to be inline with other's

metamemelord: Yeah done, thanks!

VikashLNU (approved):

Issue #654, opened 2022-08-12
Go package x/sys needs an update for darwin


viraptor:
Current version of `golang.org/x/sys` fails at compile time with:
```
# golang.org/x/sys/unix
vendor/golang.org/x/sys/unix/syscall_darwin.1_13.go:29:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_amd64.1_13.go:27:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_amd64.1_13.go:40:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_amd64.go:28:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_amd64.go:43:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_amd64.go:59:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_amd64.go:75:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_amd64.go:90:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_amd64.go:105:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_amd64.go:121:3: //go:linkname must refer to declared function or variable
vendor/golang.org/x/sys/unix/zsyscall_darwin_amd64.go:121:3: too many errors
```

After updating to current `golang.org/x/sys@v0.0.0-20220811171246-fbc7d0a398ab`, the app builds and seems to work correctly.

PR #655, opened 2022-11-14
Update golang.org/x/sys.

tjni:
This picks up a new 2022 version of golang.org/x/sys which is caused by https://github.com/golang/go/issues/49219 and is needed to fix building using Go 1.18 on aarch64-darwin. Fixes https://github.com/fnproject/cli/issues/654.

As a side note: We cannot upgrade to the latest version of golang.org/x/sys because, at least for now, it requires a minimum Go version of 1.17.


tjni:
Would it be possible to review and merge if this looks acceptable? Because the dependencies here are vendored, leading to many changes, they are not easy to patch as a package manager. In my case, I'd like to unbreak this package on Darwin for [nixpkgs](https://github.com/NixOS/nixpkgs) users.

/cc @astromechza @harryjsmith @metamemelord @JadeRedworth as the 4 most recent folks who committed to this repository. I envision many of you are quite busy, so I'll only ping occasionally.

astromechza (approved):
LGTM, builds on my x86 darwin machine too

PR #667, opened 2023-06-21
Fn cli multi arch

ssunny3141:
Removing linux arm64 from release script to fix the build failure in circile ci


rishabhsri18 (approved):
LGTM

PR #670, opened 2023-07-28
FnCLI fn deploy

Jaytee-fn:


common/common.go
```diff
@@ -512,20 +519,31 @@ func RunBuild(verbose bool, dir, imageName, dockerfile string, buildArgs []strin
...
 		if arch, ok := ShapeMap[shape]; ok {
+			var mappedArchitectures []string
 			mappedArchitectures = append(mappedArchitectures, arch...)
-			err := initializeContainerBuilder(containerEngineType, mappedArchitectures)
-			if err != nil {
-				done <- err
-				return
+			var hostedPlatform = runtime.GOARCH
+			if platform, ok := TargetPlatformMap[shape]; ok {
+				targetPlatform := strings.Join(platform," ")
```
sudhindraaoracle, line 528: Why are we appending " " to platform?

common/common.go
```diff
@@ -512,20 +519,31 @@ func RunBuild(verbose bool, dir, imageName, dockerfile string, buildArgs []strin
...
+				if targetPlatform != hostedPlatform {
+					fmt.Println("TargetedPlatform and hostPlatform are not same")
+					if containerEngineType == ContainerEngineType {
```
sudhindraaoracle, line 531: Why is this comparison needed?

sudhindraaoracle, line 531: Change the variable name to reflect docker type.
Please move this condition to the method

Jaytee-fn: done

common/common.go
Jaytee-fn: Added comment in the code

common/common.go
```diff
@@ -511,23 +536,38 @@ func RunBuild(verbose bool, dir, imageName, dockerfile string, buildArgs []strin
...
+					if err != nil {
+						done <- err
+						return
+					}
+					dockerBuildCmdArgs = buildXDockerCommand(imageName, dockerfile, buildArgs, noCache, mappedArchitectures, containerEngineType )
+					// perform cleanup
+					defer cleanupContainerBuilder(containerEngineType)
+				} else {
+					fmt.Println("TargetedPlatform and hostPlatform are same")
+					fmt.Println("1. issuePush is ", issuePush)
```
sudhindraaoracle, line 562: Do we need these issuePush logs?

common/common.go
```diff
@@ -511,23 +536,38 @@ func RunBuild(verbose bool, dir, imageName, dockerfile string, buildArgs []strin
...
 			}
-
-			dockerBuildCmdArgs = buildXDockerCommand(imageName, dockerfile, buildArgs, noCache, mappedArchitectures)
-			// perform cleanup
-			defer cleanupContainerBuilder(containerEngineType)
-		} else {
-			dockerBuildCmdArgs = buildDockerCommand(imageName, dockerfile, buildArgs, noCache)
 		}
-
+        fmt.Println("*****containerEngineType*****",containerEngineType )
```
sudhindraaoracle, line 569: Can we merge it into a single log?

common/common.go
```diff
@@ -551,6 +591,31 @@ func RunBuild(verbose bool, dir, imageName, dockerfile string, buildArgs []strin
 		fmt.Fprintln(os.Stderr)
 		return fmt.Errorf("build cancelled on signal %v", signal)
 	}
+	if containerEngineType != containerEngineTypeDocker || issuePush {
+		fmt.Println("Using Container engine", containerEngineType, "to push as --push id only for docker and --load is for podman")
```
sudhindraaoracle, line 595: Optimise these logs.

Log statement need not be true because we use push even for podman with Build command

sudhindraaoracle (commented):
Optimise the logs and format them properly. Logic seems to be fine.

PR #680, opened 2024-03-14
Fix security vulnerability

marvin659:
Fixed security vulnerability issues identified in the following packages:
GHSA-3vm4-22fp-5rfm+golang.org/x/crypto
GHSA-4374-p667-p6c8+golang.org/x/net
GHSA-69cg-p879-7622+golang.org/x/net
GHSA-69ch-w2m2-3vjp+golang.org/x/text
GHSA-83g2-8m93-v3w7+golang.org/x/net
GHSA-8c26-wmh5-6g9v+golang.org/x/crypto
GHSA-gwc9-m7rh-j2ww+golang.org/x/crypto
GHSA-ppp9-7jff-5vj2+golang.org/x/text
GHSA-vvpx-j8f3-3w6h+golang.org/x/net

Remove Ruby 2.7 support as it reaches the end of life. Without removing that, CI build will not work.


langs/base.go
```diff
@@ -55,7 +55,7 @@ func init() {
 	registerHelper(&KotlinLangHelper{})
 
 	// for older versions support backwards compatibility
-	fallBackOlderVersions["ruby"] = &RubyLangHelper{Version: "2.5"}
+	fallBackOlderVersions["ruby"] = &RubyLangHelper{Version: "3.1"}
```
siddumk, line 58: Which CI build we are talking about here?

siddumk: @ahjones is it ok to remove older version of ruby? Do we need to update any other documentation for the same.

siddumk (commented):
We would like to stash commit which is related to ruby  and Security vulnerability fix as separate commit.

siddumk (approved):

PR #693, opened 2025-02-11
Make fn-cli compatiable with Oracle Cloud Shell based on OL8

pulkitsxn059:


sivakumart (commented):
The change looks good to me @pulkitsxn059. 

My only question/suggestion is that should `config/osrelease.go` be exhaustive as it is now? If we only intend to use `Release.Name` and `Release.PlatformID` could we simplify `osrelease.go` to only scan for those? This would keep the change minimal and focussed for what we need in the PR.

Another minor nit: Instead of populating global variables from Parse, we can have Parse return a struct.

sivakumart (approved):
Thanks for addressing the earlier comments. LGTM.

sivakumart (approved):

Jaytee-fn:
Can we add release notes as well for whats added in this version on FnCLI?

snehaoracle (approved):

PR #706, opened 2026-03-12
local debug support for OCI Functions development

marvin659:
### Tests done
- Test with Go function with --local-debug
- Test with Java function with --local-debug
- Test with Python  with --local-debug (for python functions that use requirements.txt or setup.py)


langs/python.go
```diff
@@ -91,23 +91,28 @@ func (h *PythonLangHelper) Entrypoint() (string, error) {
...
-		r = append(r, fmt.Sprintf(`
-			%v -r requirements.txt &&\
+		r = append(r, fmt.Sprintf(
+			`%v -r requirements.txt &&\
 			    rm -fr ~/.cache/pip /tmp* requirements.txt func.yaml Dockerfile .venv &&\
 			    chmod -R o+r /python`, pip_cmd))
 	}
 	r = append(r, "ADD . /function/")
 	if exists("setup.py") {
-		r = append(r, "python setup.py install")
```
marvin659, line 110: Reason for the change:
"python setup.py install" should be replaced by "python -m pip install ."

https://packaging.python.org/en/latest/discussions/setup-py-deprecated/

common/common.go
```diff
@@ -1240,3 +1253,69 @@ func untarStream(r io.Reader) error {
 		}
 	}
 }
+
+func PullImage(image string) error {
```
marvin659, line 1257: This function was in objects/server/server.go. Moving it to common.

pdenness (approved):

PR #710, opened 2026-04-21
Add code-only build support

Jaytee-fn:
The change adds local build support for code-only functions.

This PR introduces a code-only build path that packages local function source into a zip archive instead of performing the traditional container image build flow.

The PR -
- detect 'code_only: true' in 'fn build'
- validate required local tooling before packaging
- create a local archive named:
  - <function-name>.<version>.zip
- bypass the legacy image/container-engine build path for code-only functions
- apply runtime-specific archive layout behaviour for code-only packaging

Validation

Build a code-only function
fn build --working-dir /path/to/code-only-function

### DIFF
--- a/commands/build.go
+++ b/commands/build.go
@@ -17,11 +17,16 @@
 package commands
 
 import (
+	"archive/zip"
 	"fmt"
 	"github.com/fnproject/cli/common"
 	"github.com/urfave/cli"
+	"io"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"sort"
+	"strings"
 )
 
 // BuildCommand returns build cli.command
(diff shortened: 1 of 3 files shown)


commands/build.go
```diff
@@ -122,3 +139,178 @@ func (b *buildcmd) build(c *cli.Context) error {
...
+}
+
+func validateCodeOnlyBuildTooling(ff *common.FuncFileV20180708) error {
+	runtimeName := ""
+	if ff.Runtime_config != nil {
+		runtimeName = strings.TrimSpace(ff.Runtime_config.Runtime_name)
+	}
+	baseRuntime := codeOnlyBaseRuntime(runtimeName)
+	switch {
+	case strings.HasPrefix(baseRuntime, "java"), strings.HasPrefix(baseRuntime, "kotlin"):
```
neelpandya, line 161: No kotlin and ruby support. pls remove.

PR #714, opened 2026-04-24
Hot Reload feature for OCI Functions local development with Fn Server

marvin659:
Hot Reload feature for OCI Functions local development with Fn Server

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,26 @@ curl -LSs https://raw.githubusercontent.com/fnproject/cli/master/install | sh
 ## CLI Development
 * Refer to the [Fn CLI Wiki](https://github.com/fnproject/cli/wiki) for development details.
 
+## Watch (local auto-deploy)
+To watch a directory and automatically redeploy to a local Fn server when files change:
+
+```sh
+fn watch --app <app>
+```
+
+This watches the current directory recursively and triggers:
+
+```sh
+fn deploy --app <app> --local
+```
+
+### Ignoring paths
+`fn watch` ignores these directories by default:
+
+- `.git`, `.fn`, `node_modules`, `target`, `dist`, `vendor`, `Dockerfile-fn-temp*`
+
+You can add more ignore rules by creating a `.fnignore` file in the watched directory (one pattern per line; `#` comments supported), and/or by passing `--ignore` flags.
+
 ### Build from source
 See [CONTRIBUTING](https://github.com/fnproject/cli/blob/master/CONTRIBUTING.md) for instructions to build the CLI from source.
 
--- a/README.md
+++ b/README.md
@@ -35,19 +35,18 @@ fn watch --app <app>
 This watches the current directory recursively and triggers:
 
 ```sh
-fn deploy --app <app> --local
+fn deploy --app <app> --local --no-bump
 ```
 
 ### Ignoring paths
 `fn watch` ignores these directories by default:
 
-- `.git`, `.fn`, `node_modules`, `target`, `dist`, `vendor`, `Dockerfile-fn-temp*`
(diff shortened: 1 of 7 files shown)


common/common.go
```diff
@@ -738,7 +738,7 @@ func writeTmpDockerfile(helper langs.LangHelper, dir string, ff *FuncFile) (stri
 		return "", errors.New("entrypoint and cmd are missing, you must provide one or the other")
 	}
 
-	fd, err := ioutil.TempFile(dir, "Dockerfile")
+	fd, err := ioutil.TempFile(dir, "Dockerfile-fn-tmp")
```
marvin659, line 741: changed the temp file name to be more specific so that "fn watch" will ignore this file if it got changed by "fn deploy"

commands/watch.go
```diff
@@ -0,0 +1,333 @@
...
+			if !ok {
+				return nil
+			}
+			fmt.Fprintf(os.Stderr, "watch error: %v\n", err)
+		}
+	}
+}
+
+func runFnDeployLocal(ctx context.Context, dir string, appName string) error {
+	cmd := exec.CommandContext(ctx, "fn", "deploy", "--app", appName, "--local", "--no-bump")
```
Jaytee-fn: Looks like this step depends on the fn binary available on PATH. It may not be same binary user used to start watch. Would it be safe to use os.Executable() so watch always use the same exact CLI instance that started it?

commands/watch.go
```diff
@@ -0,0 +1,333 @@
...
+				Name:  "app",
+				Usage: "Local app name to deploy to",
+			},
+			cli.DurationFlag{
+				Name:        "debounce",
+				Usage:       "Debounce duration before triggering deploy (e.g. 500ms, 2s)",
+				Value:       defaultWatchDebounce,
+				Destination: &cmd.debounce,
+			},
+			cli.StringSliceFlag{
```
Jaytee-fn, line 67: The help text says --ignore matches path segments it also gives --ignore '*.log' as an example, but the implementation only check
exact segment equality (s == p), or
filepath.Match

we could have some issue here for the nested log files.

README.md
```diff
@@ -25,6 +25,26 @@ curl -LSs https://raw.githubusercontent.com/fnproject/cli/master/install | sh
...
+This watches the current directory recursively and triggers:
+
+```sh
+fn deploy --app <app> --local
+```
+
+### Ignoring paths
+`fn watch` ignores these directories by default:
+
+- `.git`, `.fn`, `node_modules`, `target`, `dist`, `vendor`, `Dockerfile-fn-temp*`
```
Jaytee-fn: it should be Dockerfile-fn-tmp*

README.md
```diff
@@ -25,6 +25,26 @@ curl -LSs https://raw.githubusercontent.com/fnproject/cli/master/install | sh
...
+To watch a directory and automatically redeploy to a local Fn server when files change:
+
+```sh
+fn watch --app <app>
+```
+
+This watches the current directory recursively and triggers:
+
+```sh
+fn deploy --app <app> --local
```
Jaytee-fn: should we also add --no-bump here?
fn deploy --app <app> --local --no-bump

commands/watch.go
marvin659: ✅

commands/watch.go
marvin659: ✅  fixed and added test case

README.md
marvin659: ✅

README.md
marvin659: ✅

Jaytee-fn (approved):

PR #717, opened 2026-05-06
Add Node 24 and Java 21 runtime support

Jaytee-fn:
The change is to -
- add support for node24 and java21 in Fn CLI
- update default runtimes so 'node' resolves to Node 24 and 'java' resolves to Java 21
- keep legacy fallback behaviour for older versionless func.yaml runtimes (node -> node22, java -> java17) during image stamping
- extend tests for runtime defaults, fallback behaviour, and image generation


Preeti-Misra (approved):
changes look good to me

PR #718, opened 2026-05-07
Add long-running OCI Functions support to Fn CLI

Jaytee-fn:
This PR adds long running functions support to Fn CLI. It adds detached timeout, success/failure destination, detached invoke type support.

The PR includes CLI support for -
--detached-timeout
--on-success
--on-failure
--clear-on-success
--clear-on-failure
- Adds `func.yaml` changes for OCI detached mode settings
- Adds deploy/update handling so detached mode settings are applied from func.yaml
- Adds inspect/list/output helpers for detached mode views
- Adds OCI shim mappings for detached timeout and success/failure destinations
- Adds warnings when OCI-only options are used with non-Oracle/local providers

This PR also upgrades the OCI Go SDK and project toolchain because detached mode support is only available in newer OCI SDK versions that require Go 1.24+.


marvin659 (approved):
LGTM

PR #719, opened 2026-05-11
fix hot relad issue

marvin659:
We could not use "no-bump" option in deploying functions locally as the hot container running in FnServer will not detect the code change and will still use the old image.

FnServer calculated the slot hash by using image version only (and also other attributes). If we do not update image version, the FnServer will not pick up the new change if there is a hot container is already running.

Also fixed a hot container init timeout issue if developer is using local debug mode. Setting it to 5 mins so that developer has enough time to attach an IDE debugger when container is waiting to start.

### DIFF
--- a/commands/watch.go
+++ b/commands/watch.go
@@ -21,10 +21,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/fnproject/cli/common"
+	"io"
 	"os"
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"sync"
 	"syscall"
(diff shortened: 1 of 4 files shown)


commands/watch.go
```diff
@@ -140,6 +144,11 @@ func (w *watchcmd) watchLoop(ctx context.Context, root string, appName string, w
 		debounceTimer  *time.Timer
 	)
 
+	// If func.yaml changes only by its "version:" line, ignore it (those changes are
+	// frequently made automatically by tools and do not affect the deploy outcome).
+	funcYamlPath := filepath.Join(root, "func.yaml")
```
Jaytee-fn: should we resolve the actual funcfile path via the existing funcfile lookup logic instead of hard-coding 'func.yaml' ?

marvin659: yes you are right. Fixed.

Jaytee-fn (approved):

PR #720, opened 2026-05-12
fix bug in generating java fdk docker file which caused 502s error

marvin659:
There is a bug in using Fn CLI version from 0.6.49 to 0.6.55 that if you use Fn CLI to build java functions, you will experience 502s container initialization error when invoking the functions. It was a bug introduced in version 0.6.49 which caused the docker file for Java malformed.

Tested for Java/Go/Python/DotNet/Ruby/Node. Here is the test result for java
```
Updating function java-test using image java-test:0.0.2...
Successfully created function: java-test with java-test:0.0.2
➜  java-test fn -v invoke myapp java-test
Hello, world!
```


Jaytee-fn (approved):
Looks good

PR #723, opened 2026-05-18
Add PBF catalog commands, func.yaml, and deploy support to Fn CLI

Jaytee-fn:
This PR extends Fn CLI with a broader OCI Pre-Built Function (PBF) workflow, including:
- PBF catalog browse commands
- PBF-backed function create improvements
- persisted PBF source details in 'func.yaml'
- 'fn init' and 'fn deploy' support for PBF-backed functions
- inspect/list visibility for PBF-backed functions

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,42 @@ deploy:
 
 When these OCI-specific flags are used with a non-Oracle provider or local Fn server workflows, Fn CLI accepts them and emits user-friendly warnings where the settings are not applicable.
 
-### Pre-Built Function create support
+### Pre-Built Function (PBF) support
+List available Pre-Built Functions:
+
+```sh
+fn list pbfs
+fn list pbfs --search Document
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -151,7 +151,42 @@ deploy:
...
+
+```sh
+fn get pbfs <pbf-name-or-ocid>
+```
+
+List versions for a PBF:
+
+```sh
+fn list pbfs versions <pbf-name-or-ocid>
+fn list pbfs versions <pbf-name-or-ocid> --latest
```
neelpandya: imo , it should be --current instead of --latest . See, we can rollback to any old version , thus nullifying the "latest" construct.

Jaytee-fn: changed the flag from --latest to --current

neelpandya (approved):
Looks good.

stuartjturner (approved):

PR #727, opened 2026-07-09
Cache invoke endpoints for fn invoke

Jaytee-fn:
This PR adds endpoint caching for 'fn invoke' to reduce unnecessary control-plane calls during repeated function invocations through Fn CLI.

Previously, each 'fn invoke <app> <function>' resolved the function invoke endpoint by looking up the app and function every time. For OCI Functions, that resulted in repeated 'ListApps' / 'ListFns' calls before every invoke. Customers using Fn CLI in frequent invocation loops could generate high CP traffic and hit throttling.

Changes

- Added a persistent invoke endpoint cache at '~/.fn/invoke-endpoint-cache.json'.
- Cache key includes:
  - current context
  - provider
  - API URL
  - compartment ID
  - app name
  - function name
- Added default cache TTL of '15m'.
- Added controls to disable endpoint caching for a single invoke or tune the cache TTL through a CLI flag or environment variable-
  - --no-endpoint-cache
  - --endpoint-cache-ttl
  - FN_INVOKE_ENDPOINT_CACHE_TTL
- Preserved existing '--endpoint' behaviour as a direct bypass.
- Added cache invalidation after:
  - fn deploy
  - fn update function
  - fn delete function
  - fn delete app
- Cache access is guarded for concurrent CLI invocations, and writes are performed safely via temporary-file replacement.

Manual verification:

- First fn invoke <app> <function> creates ~/.fn/invoke-endpoint-cache.json.
- Second invoke for the same app/function uses the cached endpoint.
- --no-endpoint-cache bypasses the cache.
- Successful deploy/update/delete invalidates matching cache entries.

### DIFF
--- a/common/invoke_endpoint_cache.go
+++ b/common/invoke_endpoint_cache.go
@@ -1,0 +1,283 @@
+/*
+ * Copyright (c) 2019, 2020 Oracle and/or its affiliates. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package common
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/url"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/fnproject/cli/config"
+	"github.com/fnproject/fn_go/provider"
+	"github.com/fnproject/fn_go/provider/oracle"
+	"github.com/gofrs/flock"
+	"github.com/spf13/viper"
+)
+
+const (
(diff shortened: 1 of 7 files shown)


marvin659 (approved):

common/invoke_endpoint_cache.go
```diff
@@ -0,0 +1,289 @@
...
+
+// Get returns a cached endpoint when it exists and has not expired.
+func (c *InvokeEndpointCache) Get(key InvokeEndpointCacheKey, ttl time.Duration) (string, bool, error) {
+	if ttl <= 0 {
+		return "", false, nil
+	}
+
+	var endpoint string
+	var ok bool
+	err := c.withLockedData(false, func(data *invokeEndpointCacheData) error {
```
marvin659: actually could we avoid the lock when reading the file? I think that would force customer who has multiple Fn CLI processes to run that part of code sequentially during fn invocation.

marvin659: p.s. frequency of write should be low. Even there is a contention, the read could fail safe and query the CP endpoint in case if the read operation failed. What do you think?

marvin659:
Approved despite the above concern, it is an improvement to the original implementation anyway.

common/invoke_endpoint_cache.go
Jaytee-fn: Good point. I agree that reads do not need to take the file lock since cache writes are done via temp-file replacement. A lockless read can safely treat any read/parse error as a cache miss and fall back to resolving the endpoint through CP.

I will keep locking around writes/invalidation because those paths do read-modify-write and need protection against lost updates, but I can remove the lock from cache reads to avoid serializing concurrent 'fn invoke' processes on cache hits.

Jaytee-fn: Updated

PR #732, opened 2026-09-03
Support Cloud Shell on Oracle Linux 9

Jaytee-fn:
Recognize both OL8 and OL( clod shell env
Retain cloud shell docker/podman compatability path on OL9
Make the cross-arch error message Cloud shell version neutral
Update the /etc/os-release test fixture to OL9
Add test coverage for OL8 and OL9 cloud shell detection

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -52,17 +52,18 @@ const (
 	EnvFnRegistry = "registry"
 	EnvFnContext  = "context"
 
-	OCI_CLI_AUTH_ENV_VAR                  = "OCI_CLI_AUTH"
-	OCI_CLI_CLOUDSHELL_ENV_VAR            = "OCI_CLI_CLOUD_SHELL"
-	OCI_CLOUDSHELL_OS_NAME                = "Oracle Linux Server"
-	OCI_CLOUDSHELL_OL8_VERSION_IDENTIFIER = "el8"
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -209,23 +210,28 @@ func ensureConfiguration() error {
...
 
 	return nil
 }
 
+func isCloudShellOS(osrelease *OSRelease) bool {
+	if osrelease.Name != OCI_CLOUDSHELL_OS_NAME {
+		return false
+	}
+
+	return osrelease.PlatformID == OCI_CLOUDSHELL_OL8_PLATFORM_ID ||
```
pulkitsxn059, line 231: Just a thought, we are doing equals it is fine. But if there is some change in those identifier this condition might fail. Although I believe this should not change. But do we think it would worth while to have an contains condition here? Just like we had for OL8.

Jaytee-fn: exact PLATFORM_ID match would only allow supported and validated cloud shell platform identifiers. 
A contains check could unintentionally enable cloud shell compatability path for an unvalidated platform. The previous Contains check existed because the code matched only ol8 substring . With explicit OL8 and OL9 support I think exact matching is clearer and safer as compatibility path also changes Docker/Podman behaviour and cross architecture builds

pulkitsxn059: ok makes sense. I can't approve the PR, but I am good with this change.

stuartjturner (approved):

PR #735, opened 2026-09-17
Preserve function settings and fix Windows context switching

Jaytee-fn:
PR fixes two Fn CLI issues -

- Preserve existing function config when fn update function is run without --config
- Fix windows failures in fn use context <context> when replacing ./fn/config.yaml


pulkitsxn059:
Reviewed the PR and I am good with the changes.

stuartjturner (approved):

